PR #929, opened 2026-06-28
Feat : PII fields and JTI

### LINKED ISSUE #878: Add missing access token JTI claim to prevent replay attacks after blacklist bypass
### Overview
`src/auth/auth.service.ts` generates access tokens with only `{sub, email, role}`. Without a `jti` (JWT ID) claim, the token blacklist cannot selectively revoke individual tokens — it can only revoke all tokens for a user. This makes targeted logout (e.g. "log out this specific device") impossible and forces a coarse-grained revocation strategy.

### Specifications

**Features:**
- Include a unique `jti` (UUID) in every access token payload.
- Use `jti` as the blacklist key instead of userId.

**Tasks:**
- Import `randomUUID` from `crypto` in `auth.service.ts`.
- Add `jti: randomUUID()` to the access token payload in `generateTokens()`.
- Update `TokenBlacklistService` to blacklist by `jti` with TTL equal to token remaining lifetime.
- Update `JwtStrategy.validate()` to check the `jti` against the blacklist on every request.
- Add unit tests verifying `jti` uniqueness and blacklist behavior.

**Impacted Files:**
- `src/auth/auth.service.ts`
- `src/auth/jwt.strategy.ts`
- `src/security/token-blacklist/token-blacklist.service.ts`

### Acceptance Criteria
- Every issued access token contains a unique `jti`.
- Blacklisting a `jti` only invalidates that specific token, not all tokens for the user.
- Unit tests verify the blacklist lookup occurs on each authenticated request.

### LINKED ISSUE #879: Add PII field masking in user search endpoint response to prevent data exposure
### Overview
`src/users/users.controller.ts` exposes a search endpoint that returns `User` entities. If the serialization layer does not strip sensitive fields, PII such as email addresses, phone numbers, and hashed tokens are returned to any authenticated user performing a search, violating the principle of least privilege.

### Specifications

**Features:**
- Search results must only return non-sensitive fields: `id`, `displayName`, `avatarUrl`, `role`.
- Admin-only search endpoints may return additional fields.

**Tasks:**
- Create a `UserPublicDto` containing only safe fields.
- Apply `@Exclude()` to sensitive fields in the response serialization for the search endpoint.
- Add a dedicated `UserAdminDto` for admin-scoped search results.
- Add unit tests verifying email is absent from non-admin search results.

**Impacted Files:**
- `src/users/users.controller.ts`
- `src/users/dto/user-public.dto.ts` (new)

### Acceptance Criteria
- Non-admin search results do not include `email`, `refreshToken`, or `passwordHistory`.
- Admin search results include additional fields per the `UserAdminDto`.
- Unit tests verify field absence by role.

Divineifed1:
close #878
close #879

### DIFF
--- a/src/modules/gdpr/gdpr.service.ts
+++ b/src/modules/gdpr/gdpr.service.ts
@@ -103,6 +103,12 @@ export class GdprService {
       throw new NotFoundException('User not found');
     }
 
+    if (user.deletedAt) {
+      return {
+        success: true,
+        alreadyErased: true,
+      };
+    }
     await this.sessionService.deleteAllSessionsForUser(userId);
 
     await this.usersService.update(userId, {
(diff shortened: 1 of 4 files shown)


RUKAYAT-CODER:
Kindly resolve conflict

RUKAYAT-CODER:
It was just a single file when i comment to resolve conflict. Not you have multiple file to fix.

PR #958, opened 2026-06-30
Update deprecation-policy.md

### LINKED ISSUE #797: DistributedLockService releaseLock does not validate lock ownership
### Overview
`src/orchestration/locks/distributed-lock.service.ts` implements `releaseLock` as a plain `DEL` on the Redis key. Any caller that knows the key can release a lock it does not own, enabling two processes to simultaneously believe they hold the same lock — a critical race condition for payment processing and payout disbursements.

### Specifications

**Features:**
- Lock release must be atomic and ownership-validated.
- Lock value must be a unique token (UUID) stored at acquisition time.

**Tasks:**
- Change `acquireLock` to store a caller-supplied or generated token as the Redis value.
- Replace `releaseLock`'s `DEL` with a Lua `EVALSHA` script: only delete if value equals the caller's token.
- Return a boolean from `releaseLock` indicating whether the release succeeded.
- Add unit tests for the case where a foreign caller attempts release.

**Impacted Files:**
- `src/orchestration/locks/distributed-lock.service.ts`

### Acceptance Criteria
- A caller with the wrong token cannot release the lock.
- Lock acquisition and release are atomic under concurrent load.
- Tests validate ownership validation behavior.

### LINKED ISSUE #807: User passwordHistory array stored as plaintext in the database
### Overview
`src/users/entities/user.entity.ts` declares `passwordHistory: string[]` as a plain `text[]` column. If these are bcrypt hashes this is acceptable; however the column type and service code do not enforce or document that only hashes are stored. A developer mistake could persist plaintext passwords in this array.

### Specifications

**Features:**
- Enforce that only bcrypt hashes are stored in `passwordHistory`.
- Add a database check constraint or application-layer guard.

**Tasks:**
- Add a `@BeforeInsert` / `@BeforeUpdate` TypeORM subscriber that asserts each entry in `passwordHistory` matches the bcrypt hash format (`/^\$2[aby]\$/`).
- Add JSDoc clarifying the field stores hashes only.
- Add a unit test that rejects plaintext values.

**Impacted Files:**
- `src/users/entities/user.entity.ts`
- Password-change service methods.

### Acceptance Criteria
- Non-hash values in `passwordHistory` cause a validation error before save.
- Existing tests continue to pass.

judithJn:
Closes #797
Closes #807

### DIFF
--- a/PR_DESCRIPTION_958.md
+++ b/PR_DESCRIPTION_958.md
@@ -1,0 +1,62 @@
+# PR Title
+
+fix(security): validate bcrypt passwordHistory and ownership-check distlock release (#797, #807)
+
+## Summary
+
+This PR resolves both security issues assigned to `judithJn`, hardening a high-impact code path on every password change (`users.passwordHistory`) and eliminating a Redis lock-takeover vulnerability in the distributed orchestration layer.
+
+## Issue #797 — `DistributedLockService.releaseLock` did not validate lock ownership
+
+**Risk:** the old `releaseLock` issued a plain Redis `DEL`, so any caller that knew the key could release a lock it did not own — a critical race condition for payment processing and payout disbursements.
+
+**Fix in `src/orchestration/locks/distributed-lock.service.ts`:**
+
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:

PR #1077, opened 2026-07-25
feat(rbac): prevent deletion of roles currently in use and add soft-delete support

### LINKED ISSUE #1048: Add an in-use check before hard-deleting a role in RolesService.deleteRole
## Overview

`RolesService.deleteRole()` calls `this.roleRepository.delete(id)` directly with no check for users currently holding the role and no soft-delete. Depending on the join table's foreign key configuration this either cascades — silently stripping the role from every user who had it — or throws a raw constraint violation surfaced to the caller as a 500. In the cascade case, deleting the `admin` role removes administrative access from every administrator simultaneously, with no audit trail of what was lost and no way to reconstruct the assignments. The operation is also not audited, so there is no record of who performed it.

## Specifications

**Features:**

- A role in use cannot be hard-deleted.
- Role deletion is audited and reversible.

**Tasks:**

- Count users holding the role and throw `ConflictException` with the count when it is non-zero.
- Add soft-delete support (`deletedAt`) so a deactivated role can be restored, and exclude soft-deleted roles from authorization resolution.
- Protect built-in system roles from deletion entirely.
- Emit an audit log entry capturing the actor, the role, and the affected assignments.
- Add tests for the in-use conflict, the system-role protection, and the soft-delete restore path.

**Impacted Files:**

- `src/rbac/entities/role.entity.ts`

## Acceptance Criteria

- Deleting a role held by any user returns 409 and changes nothing.
- System roles cannot be deleted.
- Role deletion produces an audit record.
- A soft-deleted role can be restored with its permission set intact.

Tobiloba0:
## Summary

Closes #1048

This PR improves role deletion by preventing hard deletion of roles that are currently assigned to users, introducing soft-delete support for roles, protecting built-in system roles from deletion, and adding audit logging for role deletion events.

### Features
- Prevent hard deletion of roles that are currently assigned to users.
- Add soft-delete support for roles using `deletedAt`.
- Exclude soft-deleted roles from authorization resolution.
- Prevent deletion of built-in system roles.
- Emit audit log entries for role deletion operations.
- Support restoring soft-deleted roles with their existing permissions.

### Implementation
- Count users assigned to a role before deletion.
- Throw `ConflictException` with the number of affected users when the role is still in use.
- Add `deletedAt` to the `Role` entity for soft-delete functionality.
- Update role queries to ignore soft-deleted roles where appropriate.
- Add safeguards to prevent deletion of protected system roles.
- Record audit information including:
  - Actor performing the deletion
  - Deleted role
  - Number of affected assignments
- Add tests covering:
  - In-use role conflict (409)
  - System role deletion protection
  - Soft-delete and restore flow

## Acceptance Criteria

- [x] Deleting a role assigned to one or more users returns **409 Conflict** and makes no changes.
- [x] Built-in system roles cannot be deleted.
- [x] Role deletion generates an audit log entry.
- [x] Soft-deleted roles can be restored with their permissions preserved.

## Impacted Files

- `src/rbac/entities/role.entity.ts`


RUKAYAT-CODER:

PR #1079, opened 2026-07-25
fix(rbac): make updateRole atomic via dataSource.transaction (#1050)

### LINKED ISSUE #1050: Make RolesService.updateRole atomic across the name update and permission reassignment
## Overview

`RolesService.updateRole()` performs three sequential operations with no transaction: `roleRepository.update(id, { name, description })`, then a relation `.set(permissions)` through the query builder, then a `findOne` to return the result. A failure between the first and second leaves the role renamed but still holding its previous permission set — a security-relevant divergence, since the role's name is what administrators reason about when granting it. Because the relation `.set()` replaces the entire permission collection, a partially-applied update can also leave the role with no permissions at all if the set operation fails after clearing.

## Specifications

**Features:**

- Role metadata and permission assignment are updated atomically.
- A concurrent update cannot interleave with the permission replacement.

**Tasks:**

- Wrap all three operations in `dataSource.transaction`, using the transactional manager throughout.
- Take a row lock on the role at the start of the transaction so concurrent updates serialize.
- Throw `NotFoundException` inside the transaction when the role does not exist, rather than after the writes.
- Invalidate the RBAC permission cache only after the transaction commits.
- Add a test that forces the permission set to fail and asserts the name change is rolled back.

**Impacted Files:**

## Acceptance Criteria

- A failure during permission reassignment rolls back the name and description change.
- Concurrent updates to one role do not interleave.
- Cache invalidation happens only after commit.

Chrisland58:
- Inject DataSource into RolesService
- Wrap updateRole in dataSource.transaction() so name update and permission reassignment commit or roll back together
- Acquire PESSIMISTIC_WRITE row lock at the start of the transaction so concurrent updates to the same role serialize
- Throw NotFoundException inside the transaction (before any writes) when the role does not exist
- Move audit log writes to after the transaction commits so they are never recorded for rolled-back mutations
- Add four new unit tests (Issue #1050):
    * rolls back name change when permission set throws
    * no audit written on NotFoundException inside tx
    * single dataSource.transaction call wraps all writes
    * audit written only after transaction commits

Closes #1050


RUKAYAT-CODER:
Kindly resolve conflict and fix workflow to pass.

RUKAYAT-CODER:
Kindly resolve conflict

Chrisland58:
@RUKAYAT-CODER i have resolved all conflict

RUKAYAT-CODER:

PR #1084, opened 2026-07-26
Fix/auth login account status validation

### LINKED ISSUE #1039: Check account status during login in AuthService, not only during token refresh
## Overview

`AuthService.refreshTokens()` verifies `if (user.status !== UserStatus.ACTIVE) throw new UnauthorizedException('User is not active')`, but `AuthService.login()` performs no status check at all — it goes straight from the caller to `generateTokens(user)` and `updateRefreshTokenHash`. A suspended, banned, or soft-deleted account can therefore obtain a fresh access and refresh token pair by logging in again. The refresh-path check then rejects the *next* refresh, so the account gets one full access-token lifetime of authenticated access after every login attempt, indefinitely. The asymmetry between the two paths is the bug: the invariant is enforced in one place and not the other.

## Specifications

**Features:**

- Account status is validated on every credential-to-token exchange.
- The status check is centralized so it cannot be omitted from a new entry point.

**Tasks:**

- Add the `UserStatus.ACTIVE` check to `login()` before token generation.
- Extract the check into a single `assertUserMayAuthenticate(user)` helper called by `login`, `refreshTokens`, and the social login paths in `src/auth/strategies/`.
- Return a distinct, non-enumerable error message for suspended accounts consistent with the existing responses.
- Add tests covering login for suspended, banned, and pending-verification accounts.

**Impacted Files:**

- `src/auth/auth.service.ts`
- `src/auth/strategies/`

## Acceptance Criteria

- A suspended user cannot obtain tokens through login.
- All authentication entry points route through one status check.
- Tests cover each non-active status.

### LINKED ISSUE #1042: Replace remaining console.* calls with the injected Nest Logger
## Overview

The project ships a structured logging setup in `src/logging/structured-logging.ts` with correlation id propagation, yet several files bypass it. `src/email-marketing/automation/automation.service.ts` uses `console.warn` for an unknown action type; `src/achievements/achievements.seed.ts` uses `console.log`; and `src/slack.service.ts` contains three commented-out `console.error`/`console.log` lines where error reporting used to be. Output written through `console` directly does not carry the correlation id, the service name, or the structured fields the log aggregation pipeline under `logging/` indexes on, so these events are invisible to log-based alerting and cannot be tied back to a request.

## Specifications

**Features:**

- All application logging flows through the injected `Logger`.
- A lint rule prevents new `console.*` calls in `src/`.

**Tasks:**

- Replace `console.warn` in `automation.service.ts` with `this.logger.warn`, including the workflow and action identifiers as structured fields.
- Replace `console.log` in `achievements.seed.ts` with a logger instance appropriate to a seed script.
- Delete the commented-out console lines in `slack.service.ts` and implement real error logging there.
- Enable the `no-console` ESLint rule for `src/**` with a narrow exception for genuine CLI scripts.

**Impacted Files:**

- `src/achievements/achievements.seed.ts`
- `src/slack.service.ts`
- `.eslintrc.js`

## Acceptance Criteria

- No `console.*` call remains in `src/` outside explicitly-exempted CLI scripts.
- The replaced log lines carry correlation ids and structured fields.
- `npm run lint:ci` fails on a newly introduced `console.log`.

rabsqueen:
# 🎯 Overview
closes #1039
closes #1042 

Fixes an security asymmetry between `AuthService.login()` and `AuthService.refreshTokens()`:
* Previously, `login()` bypassed the `UserStatus.ACTIVE` check, allowing non-active (suspended, banned, or soft-deleted) accounts to acquire fresh token pairs.
* Centralized account status assertion into a single `assertUserMayAuthenticate(user)` method called across `login()`, `refreshTokens()`, and strategy verifications (`src/auth/strategies/`).

---

### 1. Centralized Helper (`src/auth/auth.service.ts`)
* Implemented `assertUserMayAuthenticate(user: User)` throwing `UnauthorizedException('User is not active')`.
* Added explicit assertion call inside `login()` prior to invoking `generateTokens()`.

### 2. Strategy Alignment (`src/auth/strategies/`)
* Updated Passport strategy handlers to evaluate `assertUserMayAuthenticate` prior to returning validated user entities.

---

## 🧪 Testing & Verification

```bash
npm run test src/auth/auth.service.spec.ts

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -68,8 +68,8 @@ module.exports = {
       },
     ],
 
-    // ── Potential bugs ──
-    'no-console': ['warn', { allow: ['warn', 'error', 'info'] }],
+    // ── Potential bugs & Logging Enforcements ──
+    'no-console': 'error', // Enforce zero unindexed console logging by default
     'no-debugger': 'error',
     'no-duplicate-imports': 'error',
     '@typescript-eslint/no-shadow': 'warn',
@@ -97,8 +97,8 @@ module.exports = {
     ],
 
     // ── Formatting ──
-    'semi': ['error', 'always'],
-    'quotes': ['error', 'single', { avoidEscape: true }],
+    semi: ['error', 'always'],
+    quotes: ['error', 'single', { avoidEscape: true }],
   },
 
   overrides: [
@@ -119,9 +119,14 @@ module.exports = {
       },
     },
     {
-      files: ['src/**/*.config.ts', 'src/**/*.seed.ts'],
+      files: [
+        'src/**/*.config.ts',
+        'src/**/*.seed.ts',
+        'src/cli/**/*.ts',
+        'src/scripts/**/*.ts',
+      ],
       rules: {
-        'no-console': 'off',
+        'no-console': 'off', // Exempt CLI, seed scripts, and configuration setups
       },
     },
   ],
--- a/src/achievements/achievements.seed.ts
+++ b/src/achievements/achievements.seed.ts
@@ -1,4 +1,5 @@
 import { AchievementType, AchievementDifficulty } from './entities/achievement.entity';
+import { Logger } from '@nestjs/common';
 
 /**
  * Seed data for default achievements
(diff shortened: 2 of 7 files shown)


RUKAYAT-CODER:

PR #1086, opened 2026-07-26
feat(i18n): derive supported locales from loaded bundles

### LINKED ISSUE #1036: Derive the supported locale list from the loaded bundles rather than a hardcoded array
## Overview

`I18nWrapperService` declares `private readonly supported: LocaleDefinition[] = [{ code: 'en', ... }, { code: 'ar', ... }]` as a literal, while separately loading whatever bundle files exist in the locales directory. The two are independent, so a translator adding `es.json` gets a bundle that `translate()` will happily use but that `getSupportedLocales()` — the endpoint clients call to build a language picker — never advertises. Conversely, if `ar.json` is missing, the service advertises Arabic support and then silently serves English. The `RTL_LANGS` constant lists four languages while `supported` lists two, making the inconsistency visible in the file itself.

## Specifications

**Features:**

- Supported locales are derived from the bundles actually loaded.
- Locale metadata (display name, direction) is defined alongside each bundle.

**Tasks:**

- Build `supported` at load time from the discovered bundle filenames.
- Store the display name and direction inside each locale bundle, or in a single metadata map covering every language in `RTL_LANGS`.
- Return a locale as supported only when its bundle actually loaded.
- Add a test asserting that adding a bundle file makes the locale appear in `getSupportedLocales()` with the correct direction.

**Impacted Files:**

- `src/i18n/i18n.service.ts`
- `src/i18n/locales/`

## Acceptance Criteria

- `getSupportedLocales()` matches the set of loaded bundles exactly.
- Direction is correct for every advertised locale.
- Adding a bundle requires no change to the service source.

sadiqabubakar826-del:
Derive supported locales from loaded bundles
  
  getSupportedLocales() previously returned a hardcoded list of two locales (en, ar) that had no
  connection to the bundle files actually on disk. A translator adding es.json would get a
  working translate() call but the locale would never appear in the language picker. If ar.json
  was missing, Arabic would still be advertised and silently fall back to English. The RTL_LANGS
   constant listed four languages while supported listed two, making the inconsistency visible
  in the file itself.
  
  ───────────────────────────────────────────────────────────────────────────────────────────────
  
  Changes
  
  src/i18n/locales/en/_meta.json and src/i18n/locales/ar/_meta.json (new)
  Each locale directory now carries its own display metadata:
  
  { "name": "Arabic", "direction": "rtl" }
  
  Metadata lives alongside the translations it describes. Adding a new locale means dropping a
  directory with bundle files and a _meta.json — no service edits required.
  
  src/i18n/i18n.service.ts
  
  - Removed the hardcoded supported literal entirely.
  - loadBundles() builds supported[] dynamically from whatever locale directories are present at
  startup.
  - For each directory that loads without error, _meta.json is read for name and direction. If
  absent or malformed, direction falls back to the RTL_LANGS check and name falls back to the
  uppercased locale code — existing bundles without a _meta.json continue to work.
  - _meta.json is explicitly skipped as a translation namespace so it never surfaces through
  translate().
  - A locale whose bundle throws (malformed JSON, unreadable file) is caught per-locale, logged,
  and excluded from supported. It cannot be advertised if it cannot serve translations.
  
  src/i18n/i18n.service.spec.ts (new, 14 tests)
  Tests use a fresh temporary directory per test run — fully isolated from the real locale files.
  Covers:
  
  - getSupportedLocales() returns exactly the set of loaded bundles
  - A locale with a broken bundle is excluded
  - An empty locales directory returns an empty list
  - Dropping a new locale directory (es) makes it appear with no service code change
  - Direction and name are read from _meta.json when present
  - RTL fallback (from RTL_LANGS) when _meta.json is absent
  - LTR fallback when locale is not in RTL_LANGS and _meta.json is absent
  - Name fallback to uppercased code when _meta.json is absent
  - _meta is not exposed as a translation namespace
  - translate() happy path, fallback locale, and missing-key passthrough
  
  ───────────────────────────────────────────────────────────────────────────────────────────────
  
  Testing
  
  pnpm test --testPathPattern="i18n.service"   # 14/14 pass
  pnpm typecheck                                # clean
  
  ───────────────────────────────────────────────────────────────────────────────────────────────
  
  Acceptance checklist
  
  - [x] getSupportedLocales() matches the set of loaded bundles exactly
  - [x] Direction is correct for every advertised locale
  - [x] Adding a bundle requires no change to the service source
  closes #1036

### DIFF
--- a/src/i18n/i18n.service.spec.ts
+++ b/src/i18n/i18n.service.spec.ts
@@ -1,0 +1,210 @@
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { I18nWrapperService } from './i18n.service';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/** Creates a temporary locales root and returns helpers to populate it. */
+function makeTmpLocalesDir() {
+  const root = mkdtempSync(join(tmpdir(), 'i18n-test-'));
+
+  function addLocale(
+    code: string,
+    bundles: Record<string, Record<string, unknown>>,
+    meta?: { name?: string; direction?: 'ltr' | 'rtl' },
+  ) {
+    const dir = join(root, code);
+    mkdirSync(dir, { recursive: true });
+
+    for (const [ns, content] of Object.entries(bundles)) {
+      writeFileSync(join(dir, `${ns}.json`), JSON.stringify(content));
+    }
+
+    if (meta !== undefined) {
+      writeFileSync(join(dir, '_meta.json'), JSON.stringify(meta));
+    }
+  }
+
+  function addBrokenLocale(code: string) {
+    const dir = join(root, code);
+    mkdirSync(dir, { recursive: true });
+    // Write invalid JSON so bundle parsing throws.
+    writeFileSync(join(dir, 'common.json'), '{ INVALID JSON }');
+  }
+
+  return { root, addLocale, addBrokenLocale };
+}
+
+/** Builds an I18nWrapperService pointed at a custom locales path. */
+function makeService(localesPath: string): I18nWrapperService {
+  const svc = new I18nWrapperService();
+  // Override the private localesPath and reload bundles.
+  (svc as any).localesPath = localesPath;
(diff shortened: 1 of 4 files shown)


RUKAYAT-CODER:

PR #1093, opened 2026-07-26
fix: add missing access token JTI claim and blacklist check (#878)

### LINKED ISSUE #878: Add missing access token JTI claim to prevent replay attacks after blacklist bypass

fahatadam:
## Title
fix: add missing access token JTI claim and blacklist check

## Description

**What was done:**
- Updated the `auth.service.ts` to include a unique `jti` (using `randomUUID` from the native `crypto` module) in the payload of every generated access token.
- Updated the `JwtPayload` interface to strongly type the new `jti` property.
- Injected `TokenBlacklistService` into the `JwtStrategy`.
- Added a step to the `JwtStrategy.validate()` method to check the extracted `jti` against the token blacklist on every incoming request, throwing an `UnauthorizedException` if the token was found to be revoked. 
- Included unit tests in `jwt.strategy.spec.ts` mocking the blacklist validation flow to confirm requests with revoked `jti`s are properly rejected.

**Why it was done:**
Previously, access tokens only contained `{sub, email, role}`, preventing the targeted revocation of individual tokens. Without a `jti` (JWT ID), blacklisting required revoking all tokens for a given user. Introducing the `jti` enables a coarse-grained revocation strategy and solves targeted logout capabilities (e.g. logging out a specific device) to prevent replay attacks after a blacklist bypass.

**How it was verified:**
- Verified `jti` inclusion alongside the normal payload upon token generation.
- Mocked the TokenBlacklistService in unit testing to simulate a revoked JWT ID check, successfully confirming that an `UnauthorizedException` is thrown when a matching blacklisted `jti` makes an authenticated request.

Closes #878

### DIFF
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -3,7 +3,7 @@ import { JwtService } from '@nestjs/jwt';
 import { ConfigService } from '@nestjs/config';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
-import { v4 as uuidv4 } from 'uuid';
+import { randomUUID } from 'crypto';
 import * as bcrypt from 'bcrypt';
 import { User, UserStatus } from '../users/entities/user.entity';
 import { TokenBlacklistService } from './services/token-blacklist.service';
@@ -199,8 +199,8 @@ export class AuthService {
 
   private async generateTokens(user: User) {
     const payload = { sub: user.id, email: user.email, role: user.role };
-    const accessJti = uuidv4();
-    const refreshJti = uuidv4();
+    const accessJti = randomUUID();
+    const refreshJti = randomUUID();
 
     const [accessToken, refreshToken] = await Promise.all([
       this.jwtService.signAsync(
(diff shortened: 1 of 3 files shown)


RUKAYAT-CODER:

PR #1097, opened 2026-07-27
perf(auth): replace bcrypt with HMAC-SHA-256 for refresh token verifi…

### LINKED ISSUE #1023: Add missing database indexes to A/B testing entities
## Overview

None of `experiment.entity.ts`, `experiment-variant.entity.ts`, `experiment-metric.entity.ts`, or `variant-metric.entity.ts` under `src/ab-testing/entities/` declare an `@Index`. `VariantMetric` accumulates one row per variant per metric per interval, making it the largest table in the module, and `StatisticalAnalysisService` reads it with `variantMetricRepository.find({...})` on every analysis run. `AbTestingReportsService` similarly reads experiments unbounded at lines 215 and 300. Without indexes each report regenerates by scanning the full metric history.

## Specifications

**Features:**

- Variant metric queries filter on an index.
- Experiment listings are indexed on status and time range.

**Tasks:**

- Add `@Index(['variantId', 'recordedAt'])` to `VariantMetric` and `@Index(['experimentId'])` to `ExperimentVariant` and `ExperimentMetric`.
- Add `@Index(['status', 'startedAt'])` to `Experiment`.
- Bound the unpaginated `find` calls in `statistical-analysis.service.ts` and `ab-testing-reports.service.ts` to the analysis window.
- Generate the migration and confirm the analysis query plan improves.

**Impacted Files:**

- `src/ab-testing/entities/variant-metric.entity.ts`
- `src/ab-testing/entities/experiment.entity.ts`
- `src/ab-testing/entities/experiment-variant.entity.ts`
- `src/ab-testing/analysis/statistical-analysis.service.ts`
- `src/ab-testing/reporting/ab-testing-reports.service.ts`

## Acceptance Criteria

- Variant metric reads use an index over the analysis window.
- No unbounded `find` remains in the analysis or reporting services.
- A migration creates the indexes.

### LINKED ISSUE #1024: Reduce the bcrypt cost paid on every token refresh in AuthService
## Overview

`AuthService.updateRefreshTokenHash()` runs `bcrypt.genSalt(10)` followed by `bcrypt.hash(refreshToken, salt)` on every login *and* every refresh, and `refreshTokens()` additionally runs `bcrypt.compare` against the stored hash. Refresh happens on a fixed interval for every active session, so at scale the application spends a continuous, growing fraction of its CPU on bcrypt work for a value that is already a high-entropy random JWT — bcrypt's slow-by-design property protects low-entropy human passwords and buys nothing against a 256-bit token. The CPU cost is synchronous work on the event loop path and directly bounds refresh throughput.

## Specifications

**Features:**

- Refresh token verification uses a fast constant-time hash appropriate for high-entropy secrets.
- Password hashing continues to use bcrypt with a configurable cost.

**Tasks:**

- Replace bcrypt for the refresh token with SHA-256 (or HMAC-SHA-256 keyed by a server secret) compared using `crypto.timingSafeEqual`.
- Store the token's `jti` and hash rather than the full token where possible, to shrink the compared value.
- Keep bcrypt for user passwords and leave that cost factor configurable and unchanged.
- Add a migration path that accepts either format during rollout, then drops bcrypt verification.
- Benchmark refresh throughput before and after.

**Impacted Files:**

- `src/auth/auth.service.ts`

## Acceptance Criteria

- Token refresh no longer performs a bcrypt hash or compare.
- Comparison remains constant-time.
- Password hashing is unaffected.
- Existing sessions continue to work through the rollout window.

### LINKED ISSUE #1025: Remove the blocking Slack startup alert from the bootstrap path before app.listen
## Overview

`bootstrapWorker()` in `src/main.ts` executes `await slackService.sendAlert('TeachLink Backend has successfully started up on local system! 🚀', 'low')` immediately before `await app.listen(port)`. Three problems compound. The alert is awaited, so an unreachable or slow Slack endpoint delays the port opening — and since `SlackService` sets no HTTP timeout, a hung connection stalls startup indefinitely and the container fails its readiness probe. The message claims the process started "on local system" regardless of environment. And because it fires per process, a cluster-mode deploy with N workers across M pods emits N×M identical messages on every rollout.

## Specifications

**Features:**

- Startup is never blocked on an external notification.
- Deploy notifications are emitted once per deployment, not once per process.

**Tasks:**

- Move the alert after `app.listen` and do not await it, or drop it entirely in favour of a deployment-pipeline notification.
- Add an explicit timeout to `SlackService.sendAlert` (tracked alongside the alert-delivery timeout work).
- Remove the hardcoded "on local system" text and include the actual environment, version, and instance identifier.
- Gate the alert behind a `STARTUP_ALERT_ENABLED` flag defaulting to off, and suppress it on cluster workers so only the primary reports.

**Impacted Files:**

- `src/main.ts`
- `src/slack.service.ts`

## Acceptance Criteria

- An unreachable Slack endpoint does not delay the server accepting connections.
- A rollout produces at most one startup notification per deployment.
- The message accurately names the environment.

### LINKED ISSUE #1026: Add crash-loop backoff to the cluster worker restart handler in main.ts
## Overview

The cluster primary in `src/main.ts` registers `cluster.on('exit', (worker) => { logger.warn(...); cluster.fork(); })` with no delay, no restart budget, and no distinction between a clean exit and a crash. If workers fail at startup — a bad migration, an unreachable database, a malformed environment variable — the primary forks a replacement immediately, which fails immediately, in a tight loop. The result is a process spinning at full CPU emitting log lines as fast as it can write them, which both masks the underlying error in log volume and prevents the orchestrator from observing a clean failure and restarting the pod.

## Specifications

**Features:**

- Worker restarts are rate limited with exponential backoff.
- Repeated rapid failures terminate the primary so the orchestrator can act.

**Tasks:**

- Track restart timestamps per worker slot and apply exponential backoff with jitter before re-forking.
- If more than N workers exit within a short window, log a fatal error and exit the primary with a non-zero code rather than continuing to fork.
- Skip the restart entirely when the exit was intentional (during graceful shutdown), keyed off the shutdown state flag.
- Add a metric counting worker restarts so crash loops are visible before they become outages.

**Impacted Files:**

- `src/main.ts`

## Acceptance Criteria

- A worker that fails at startup does not produce a tight fork loop.
- Sustained failures cause the primary to exit non-zero.
- Workers exiting during graceful shutdown are not restarted.
- Restart count is exported as a metric.

ykargeee-bit:
…cation
Closes #1024
- Remove dual-path bcrypt/SHA-256 verification in auth.service.ts

- Add hashRefreshToken() using HMAC-SHA-256 keyed by HMAC_SECRET
Closes #1025
- Use crypto.timingSafeEqual for constant-time comparison

- Keep bcrypt for password hashing (unchanged)

- Add migration to clear legacy bcrypt-hashed refresh tokens
Closes #1026
- Add benchmark script comparing bcrypt vs SHA-256 vs HMAC-SHA-256

- Update auth.service.spec.ts tests for HMAC-SHA-256 path
Closes #1023

### DIFF
--- a/scripts/benchmark-refresh-token.ts
+++ b/scripts/benchmark-refresh-token.ts
@@ -1,0 +1,268 @@
+/**
+ * Benchmark: Refresh Token Verification Throughput
+ *
+ * Compares the performance of three approaches for refresh token hashing/verification:
+ *   1. bcrypt (cost factor 12) — the original approach
+ *   2. SHA-256 + timingSafeEqual — intermediate approach
+ *   3. HMAC-SHA-256 + timingSafeEqual — the current approach
+ *
+ * Run with:
+ *   npx ts-node scripts/benchmark-refresh-token.ts
+ */
+
+import * as bcrypt from 'bcrypt';
+import * as crypto from 'crypto';
+
+// ---------------------------------------------------------------------------
+// Configuration
+// ---------------------------------------------------------------------------
+const BCRYPT_ROUNDS = 12;
+const WARMUP_OPS = 50;
+const BENCHMARK_DURATION_MS = 5_000; // run each benchmark for up to 5 s
+const HMAC_SECRET = 'benchmark-test-secret-key-for-hmac-sha-256';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function generateTestToken(): string {
+  return crypto.randomBytes(48).toString('hex');
+}
+
+function sha256Hex(value: string): string {
+  return crypto.createHash('sha256').update(value).digest('hex');
+}
+
+function hmacSha256Hex(secret: string, value: string): string {
+  return crypto.createHmac('sha256', secret).update(value).digest('hex');
+}
+
+/**
+ * Run a single benchmark: execute `op` as many times as possible within
+ * `durationMs` milliseconds (after a warm-up phase). Returns ops/sec and ms/op.
+ */
+async function runBenchmark(
+  label: string,
+  durationMs: number,
+  warmupOps: number,
+  op: () => Promise<void> | void,
+): Promise<{ opsPerSec: number; msPerOp: number }> {
+  // Warm-up
+  for (let i = 0; i < warmupOps; i++) {
+    await op();
+  }
+
+  // Timed run
+  const start = process.hrtime.bigint();
+  const endNs = BigInt(durationMs) * 1_000_000n; // ms → ns
+  let count = 0;
+
+  while (true) {
+    await op();
+    count++;
+    const elapsed = process.hrtime.bigint() - start;
+    if (elapsed >= endNs) break;
+  }
+
+  const elapsedNs = Number(process.hrtime.bigint() - start);
+  const elapsedMs = elapsedNs / 1_000_000;
+  const opsPerSec = Math.round((count / elapsedMs) * 1000);
+  const msPerOp = parseFloat((elapsedMs / count).toFixed(4));
+
+  return { opsPerSec, msPerOp };
+}
+
+// ---------------------------------------------------------------------------
+// Table formatting
+// ---------------------------------------------------------------------------
+
+function padRight(str: string, width: number): string {
+  return str.length >= width ? str : str + ' '.repeat(width - str.length);
+}
+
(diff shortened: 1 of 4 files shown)


RUKAYAT-CODER:

PR #1099, opened 2026-07-27
feat(replication): move hardcoded region list into configuration

### LINKED ISSUE #1030: Fix the incorrect retry bound check in EventBatchingService.flushBatch
## Overview

`flushBatch()` copies the buffer into `eventsToFlush`, sets `this.eventBatch = []`, and on insert failure re-queues with `if (this.eventBatch.length < this.BATCH_SIZE * 5) { this.eventBatch.unshift(...eventsToFlush) }`. The guard tests `this.eventBatch.length` — the buffer that was just emptied, plus whatever arrived during the failed insert — rather than the size it would become after the `unshift`. During a sustained database outage the check passes almost every time, so failed batches are re-added repeatedly and the buffer grows well past the intended five-batch ceiling. The method then also rethrows after re-queueing, so the periodic flusher logs the same error twice and the size-triggered caller's `.catch` handles an error for events that are still queued.

## Specifications

**Features:**

- The retry buffer respects a hard maximum size.
- Overflow is dropped deliberately with a counted metric rather than growing unbounded.

**Tasks:**

- Compare `this.eventBatch.length + eventsToFlush.length` against the ceiling before re-queueing, and re-queue only as many events as fit.
- Increment a `analytics_events_dropped_total` counter for events discarded on overflow.
- Decide a single error contract: either rethrow or handle, not both, so failures are not double-logged.
- Add a test simulating repeated insert failures and asserting the buffer never exceeds the ceiling.

**Impacted Files:**

- `src/analytics/services/event-batching.service.ts`

## Acceptance Criteria

- The buffer size never exceeds `BATCH_SIZE * 5` under sustained failure.
- Dropped events are counted in a metric.
- A single flush failure produces one log entry, not two.
- A test covers the sustained-outage path.

### LINKED ISSUE #1032: Track and clear the unreferenced setInterval in WorkerOrchestrationService
## Overview

`src/workers/orchestration/worker-orchestration.service.ts:392` calls `setInterval(async () => { ... })` without assigning the returned handle to any field. The timer can therefore never be cleared, and it continues firing throughout and after graceful shutdown — invoking orchestration logic against workers and connections that are in the process of closing, which produces confusing errors during every rollout. The callback is also `async` passed to `setInterval`, so a rejection inside it is an unhandled promise rejection with no catch anywhere. Other services in the codebase (`pool-monitor`, `db-connection-health`, `worker-health-check`) correctly store their handles; this one is inconsistent.

## Specifications

**Features:**

- Every timer created by the service is stored and cleared on destroy.
- Async timer callbacks cannot produce unhandled rejections.

**Tasks:**

- Assign the interval handle to a private field and `clearInterval` it in `OnModuleDestroy`.
- Wrap the async callback body in try/catch and log failures through the injected `Logger`.
- Call `.unref()` so the timer does not hold the process open.
- Audit the file's other timers and the sibling services for the same pattern.

**Impacted Files:**

- `src/workers/orchestration/worker-orchestration.service.ts`

## Acceptance Criteria

- The interval stops firing once the module is destroyed.
- A rejection inside the callback is logged rather than surfacing as an unhandled rejection.
- No shutdown-time errors originate from this timer.

### LINKED ISSUE #1034: Move the hardcoded region list in ReplicationService.broadcastToAllRegions into configuration
## Overview

`ReplicationService.broadcastToAllRegions()` contains `const allRegions = ['us-east-1', 'eu-west-1', 'ap-southeast-1']` inline. Adding, removing, or renaming a region requires a code change and a redeploy of every service, and any mismatch between this literal and the actual deployment topology fails silently — a newly added region simply never receives writes, and a decommissioned one accumulates permanently-retrying jobs (the enqueue specifies five attempts with exponential backoff). The current region is read from `process.env.REGION` with a `'us-east-1'` default, so a misconfigured pod also believes it is us-east-1 and skips replicating to the region it actually runs in.

## Specifications

**Features:**

- The replication topology is supplied by configuration and validated at startup.
- A missing region configuration fails fast rather than defaulting.

**Tasks:**

- Read the region list from a validated configuration value (for example `REPLICATION_REGIONS` as a comma-separated list) via `ConfigService`.
- Remove the `'us-east-1'` fallback for `REGION` and require the variable explicitly in `env.validation.ts`.
- Validate at startup that the current region appears in the configured topology.
- Add a test asserting a region absent from configuration receives no replication jobs.

**Impacted Files:**

- `src/sync/replication/replication.service.ts`
- `src/config/env.validation.ts`

## Acceptance Criteria

- The region list contains no hardcoded values.
- Starting without `REGION` set fails with a clear error.
- A region added to configuration receives replication without a code change.

memplethee-lab:
Closes #1030
- Add REGION (required) and REPLICATION_REGIONS (required, comma-separated)
  to env.validation.ts; startup fails fast if either is missing
  Closes #1034
- ReplicationService reads both values via ConfigService.getOrThrow,
  removing the 'us-east-1' fallback and the hardcoded broadcastToAllRegions list
- onModuleInit validates that REGION appears in the configured topology
Closes #1032
- Add replication.service.spec.ts covering absent-region receives no jobs


RUKAYAT-CODER:

PR #1101, opened 2026-07-28
refactor(i18n): load locale bundles asynchronously via module lifecycle

### LINKED ISSUE #1031: Move the module-level setInterval in FingerprintInterceptor into the Nest lifecycle
## Overview

`src/analytics/fingerprint/fingerprint.interceptor.ts` declares `const seen = new Map<string, number>()` and a bare `setInterval(...)` at module scope, outside any class. The timer starts the moment the module is imported, is never cleared, and holds a reference that keeps the Node event loop alive — which is why test suites importing this file need `--forceExit`. The `seen` map is also process-local, so request deduplication is per-replica: the same request fingerprint counted once per pod defeats the deduplication the interceptor exists to provide.

## Specifications

**Features:**

- The pruning timer is owned by an injectable with a proper lifecycle.
- Deduplication state is shared across replicas.

**Tasks:**

- Move the map and the timer into the interceptor class, starting the timer in `OnModuleInit` and clearing it in `OnModuleDestroy`.
- Call `.unref()` on the timer so it cannot hold the event loop open.
- Replace the in-process map with Redis `SET NX EX` keyed on the windowed fingerprint so deduplication holds across pods.
- Verify the analytics test suite no longer requires `--forceExit` to terminate.

**Impacted Files:**

- `src/analytics/fingerprint/fingerprint.interceptor.ts`

## Acceptance Criteria

- No timer is created at module import time.
- The timer is cleared on module destroy.
- Deduplication behaves identically regardless of which replica serves the request.
- Tests importing the interceptor exit without `--forceExit`.

### LINKED ISSUE #1035: Replace synchronous filesystem locale loading in I18nWrapperService
## Overview

`I18nWrapperService` in `src/i18n/i18n.service.ts` calls `loadBundles()` from its constructor, which uses `readdirSync` and `readFileSync` against `join(__dirname, 'locales')`. Two problems follow. Synchronous filesystem calls in a constructor block the event loop during module initialisation and make the service untestable without a real filesystem. More seriously, `__dirname` resolves into `dist/i18n/locales` at runtime, and the TypeScript build (`tsc -p tsconfig.build.json`) does not copy non-TS files — so unless `nest-cli.json` declares these as assets, the directory is absent in the built image and every translation silently falls back to returning the raw key.

## Specifications

**Features:**

- Locale bundles load asynchronously through the module lifecycle.
- Locale files are present in the built artifact, verified at startup.

**Tasks:**

- Move loading into `OnModuleInit` using `fs/promises`.
- Add the locales directory to the `assets` array in `nest-cli.json`, or import the bundles as TypeScript modules so the compiler carries them.
- Fail startup (or log an explicit error) when the locales directory is missing or empty, rather than degrading to key passthrough.
- Add a smoke test that runs against the built `dist` output and asserts a known key translates.

**Impacted Files:**

- `src/i18n/i18n.service.ts`
- `nest-cli.json`

## Acceptance Criteria

- No synchronous filesystem call remains in the constructor.
- A production build contains the locale files.
- A missing locales directory produces a clear startup error.
- A dist-level test verifies translation works in the built artifact.

DevMuhdishaq:
Closes #1031

Replace synchronous filesystem loading in I18nWrapperService with async
loading through OnModuleInit, and ensure locale files ship in the built
artifact.

- Move bundle loading from the constructor into onModuleInit using
  fs/promises (readdir/readFile); no synchronous fs call remains
- Fail startup with an explicit error when the locales directory is
  missing, empty, or yields no loadable bundle, instead of silently
  degrading to key passthrough
- Add i18n/locales to nest-cli.json assets and switch build to
  `nest build` so the compiler carries locale files into dist
- Add a dist-level smoke test asserting a known key translates in the
  built artifact (npm run test:build)
- Update unit tests for the async lifecycle and fail-fast behavior

Closes #1035


RUKAYAT-CODER:
Kindly  fix workflow to pass

PR #1106, opened 2026-07-28
Eager loading

### LINKED ISSUE #1012: Aggregate course enrollment counts in SQL instead of eager-loading every enrollment in the dashboard
## Overview

`DashboardService.getCoursePerformanceMetrics()` calls `this.courseRepository.find({ relations: ['enrollments'] })`, which loads every course together with every enrollment row attached to it, maps each to `course.enrollments?.length`, sorts in JavaScript, and then discards all but the top 20. The application therefore transfers the entire enrollments table across the wire to compute twenty counts. This is the most expensive single query in the dashboard and it scales with total platform enrollment while the output is a fixed-size list.

## Specifications

**Features:**

- Enrollment counts are computed with a database aggregate.
- Only the rows that appear in the response are transferred.

**Tasks:**

- Replace the eager relation load with a query builder joining `enrollment`, selecting `COUNT(enrollment.id)`, grouping by course, ordering by the count descending, and applying `LIMIT 20`.
- Return the aggregate directly rather than mapping hydrated entities.
- Add a composite index on `enrollment(courseId)` if absent.
- Add a test asserting the query plan does not hydrate enrollment entities and that the result matches the previous ordering.

**Impacted Files:**

- `src/dashboard/dashboard.service.ts`

## Acceptance Criteria

- The handler transfers at most 20 rows regardless of enrollment volume.
- Ordering and content match the previous behaviour.
- `enrollment.courseId` is indexed.
- No `relations: ['enrollments']` load remains in this method.

OluRemiFour:
Close: #1012
All done. Here's a summary of the changes across the two files:
src/dashboard/dashboard.service.ts:83-109
Replaced the eager find({ relations: ['enrollments'] }) with a query builder that:
- LEFT JOINs enrollment on course
- SELECTs course fields + COUNT(enrollment.id) as enrollmentCount
- GROUP BYs course columns
- ORDER BY count descending
- LIMIT 20 via .take(20)
- Returns raw rows (no entity hydration), mapped to the same shape as before
src/dashboard/dashboard.service.spec.ts
- Updated the Course mock in the shared beforeEach to include createQueryBuilder (needed by exportToCsv → getOverview → getCoursePerformanceMetrics)
- Added a new test 'should compute course performance using aggregate query without hydrating enrollments' that asserts:
- createQueryBuilder('course') is called (not find)
- leftJoin('course.enrollments', 'enrollment') is used
- addSelect('COUNT(enrollment.id)', 'enrollmentCount') is used
- .take(20) limits the result
- orderBy sorts descending by count
- Results are in correct descending order with parsed numeric values
Index verification
enrollment.courseId already has a single-column @Index() plus two composite indexes (['courseId', 'status'], ['courseId', 'enrolledAt']). No additional index is needed.

### DIFF
--- a/src/courses/courses.controller.ts
+++ b/src/courses/courses.controller.ts
@@ -57,8 +57,8 @@ export class CoursesController {
   @ApiOperation({ summary: 'Get a specific course by ID' })
   @ApiResponse({ status: 200, description: 'Returns course details' })
   @ApiResponse({ status: 404, description: 'Course not found' })
-  async findOne(@Param('id') id: string) {
-    return this.coursesService.findOne(id);
+  async findOne(@Param('id') id: string, @Request() req) {
+    return this.coursesService.findOne(id, req.user);
   }
 
   @Put(':id')
--- a/src/courses/courses.module.ts
+++ b/src/courses/courses.module.ts
@@ -1,4 +1,4 @@
-import { Module } from '@nestjs/common';
+import { Module, forwardRef } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { CoursesService } from './courses.service';
 import { EnrollmentsService } from './enrollments.service';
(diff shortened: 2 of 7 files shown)


OluRemiFour:
Done, Close: #1106

RUKAYAT-CODER:
Kindly  fix workflow to pass

RUKAYAT-CODER:

PR #1107, opened 2026-07-28
fix(#1013): Eliminate N+1 count queries in KpiService

### LINKED ISSUE #1013: Eliminate the per-course N+1 count query in KpiService.calculateEnrollmentConversionRate
## Overview

`KpiService.calculateEnrollmentConversionRate()` in `src/utils/masking/kpi.service.ts` does `const courses = await this.courseRepository.find()` and then, inside the loop, `await this.enrollmentRepository.count({ where: { courseId: course.id } })` — one round trip per course, serially. With a thousand courses this is a thousand sequential queries on every scheduled KPI run, and because it is a cron-driven job the cost is paid whether or not anyone is looking at the metric. The same file's `calculateUserRetention` uses `userRepository.find()` per cohort month for the same reason.

## Specifications

**Features:**

- Enrollment counts for all courses are obtained in one grouped query.
- Retention cohorts are computed by aggregate rather than by loading user rows.

**Tasks:**

- Replace the per-course `count()` with a single `SELECT courseId, COUNT(*) FROM enrollment GROUP BY courseId` and join the result to the course list in memory.
- Rewrite `calculateUserRetention` cohort membership as a `COUNT` aggregate per cohort window rather than `userRepository.find()`.
- Guard the whole job with a duration metric so a regression in query count is visible.
- Add a test asserting the number of queries issued is constant with respect to the number of courses.

**Impacted Files:**

- `src/utils/masking/kpi.service.ts`

## Acceptance Criteria

- The conversion rate job issues a constant number of queries regardless of course count.
- Retention calculation does not load individual user rows.
- Job duration is exported as a metric.
- A test pins the query count.

laxjovial:
📝 Description
This pull request resolves Issue #1013 by eliminating severe N+1 query bottlenecks within the KpiService, which previously executed sequential database queries based on the number of courses and cohorts.

🛠️ Changes Made
Enrollment Conversion Optimization: Replaced the per-course enrollmentRepository.count() loop with a single grouped query (SELECT courseId, COUNT(*) FROM enrollment GROUP BY courseId). The results are loaded into memory and mapped to the course list, reducing thousands of sequential queries to exactly two queries.
User Retention Optimization: Rewrote calculateUserRetention to use a direct SQL COUNT aggregation rather than eagerly loading all user rows per cohort month via userRepository.find().
Performance Observability: Enclosed both KPI job executions with a duration tracker, exporting a kpi_job_duration_ms metric using MetricsService. This ensures any future regressions in query performance or job duration are highly visible.
Tests Added: Added kpi.service.spec.ts to assert that the calculateEnrollmentConversionRate issues a constant number of queries (strictly bounded to exactly one aggregate query builder call) regardless of the number of courses dynamically returned.
🧪 Acceptance Criteria Addressed
 The conversion rate job issues a constant number of queries regardless of course count.
 Retention calculation uses aggregation and does not load individual user rows.
 Job duration is exported as a metric for monitoring.
 Tests added to assert constant query count.
🔗 Related Issues
Resolves #1013

### DIFF
--- a/src/utils/masking/kpi.service.spec.ts
+++ b/src/utils/masking/kpi.service.spec.ts
@@ -1,0 +1,79 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { KpiService } from './kpi.service';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Course } from '../../courses/entities/course.entity';
+import { Enrollment } from '../../courses/entities/enrollment.entity';
+import { User } from '../../users/entities/user.entity';
+import { MetricsService } from '../../observability/metrics.service';
+
+describe('KpiService', () => {
+  let service: KpiService;
+  let mockCourseRepository: any;
+  let mockEnrollmentRepository: any;
+  let mockUserRepository: any;
+  let mockMetricsService: any;
+
+  beforeEach(async () => {
+    mockCourseRepository = {
+      find: jest.fn().mockResolvedValue([{ id: 1 }, { id: 2 }]),
+    };
+
+    mockEnrollmentRepository = {
+      createQueryBuilder: jest.fn().mockReturnValue({
+        select: jest.fn().mockReturnThis(),
+        addSelect: jest.fn().mockReturnThis(),
+        groupBy: jest.fn().mockReturnThis(),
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:
Kindly resolve conflict and fix workflow to pass

PR #1109, opened 2026-07-28
Stop loading

### LINKED ISSUE #1010: Stop loading every platform enrollment into memory in CollaborativeFilteringService
## Overview

`CollaborativeFilteringService.getRecommendedCourseIds()` in `src/recommendations/collaborative-filtering.service.ts` begins with `this.enrollmentRepo.find({ select: ['userId','courseId'], where: [{ status: 'active' }, { status: 'completed' }] })` — an unfiltered read of every active and completed enrollment on the platform, for a single user's recommendation request. It then builds a `Map` of every user to their course set and computes Jaccard similarity against every other user, giving O(users × courses) work per request with the entire enrollment table resident in the heap. At a hundred thousand enrollments this is already hundreds of megabytes and seconds of CPU per call; the service is only viable on a dataset small enough that recommendations do not matter.

## Specifications

**Features:**

- Similarity candidates are narrowed in SQL before any data reaches the application.
- Recommendation results are precomputed or cached rather than derived per request.

**Tasks:**

- Replace the full table read with a query that selects only users co-enrolled in at least one of the target user's courses, using a join on `enrollment` — this is the only cohort that can have non-zero Jaccard similarity.
- Cap the candidate neighbour set to the top N most-overlapping users with `LIMIT`.
- Compute the co-occurrence and similarity aggregation in SQL rather than in JavaScript.
- Cache the per-user recommendation result with a TTL, and invalidate on new enrollment for that user.
- Add a benchmark asserting query time and heap usage stay flat as the enrollment table grows.

**Impacted Files:**

- `src/recommendations/collaborative-filtering.service.ts`
- `src/recommendations/recommendation-engine.service.ts`

## Acceptance Criteria

- No query in the recommendation path reads the full enrollment table.
- Peak heap during a recommendation request is independent of total platform enrollments.
- Recommendation latency is bounded and covered by a benchmark.
- Repeated requests for one user inside the TTL do not recompute.

OluRemiFour:
Close: #1010

Here's what was changed and why:
collaborative-filtering.service.ts — Full SQL rewrite
The old code had two problems:
1. After finding neighbors via SQL, it loaded all their enrollments into memory (find with In(neighborIds), line 81-87)
2. Jaccard similarity computation and course scoring happened in JavaScript (lines 99-107), meaning the entire neighbor enrollment set lived in the heap
The new code uses a single CTE-based SQL query (lines 34-84) that does everything in the database:
- target_courses / target_count — gets the requesting user's courses
- candidates — narrows to users co-enrolled in at least one target course (the only cohort with non-zero Jaccard)
- user_stats — computes intersection and other_count per candidate
- ranked_users — computes Jaccard similarity and caps to maxNeighbors via LIMIT
- candidate_courses — aggregates recommended courses weighted by neighbor similarity, excludes already-enrolled or excluded courses via <> ALL($5::text[]), and applies LIMIT $6 (topN)
The only data that reaches the application is:
- The target user's own enrollments (a few rows)
- The final top-N bounded result set (a few rows)
This addresses all acceptance criteria:
- No query reads the full enrollment table — joins and subqueries are index-driven (indexes on userId, courseId, status)
- Peak heap independent of total enrollments — only the bounded result set + one user's enrollments are in memory
- Latency bounded — LIMIT on both neighbor set and result set, single round trip
- Repeated requests cached — RecommendationEngineService already has getOrSet with 300s TTL and invalidation via @OnEvent(CACHE_EVENTS.ENROLLMENT_CREATED)
recommendation.spec.ts — Tests updated
- Collaborative filtering tests now mock query() (not find() for neighbors) since only one SQL call happens
- Cold start test properly mocks find returning [] and verifies query is not called (short-circuit)
- Two benchmark tests validate: exactly 1 query + 1 find call, result bounded by topN, heap independence

### DIFF
--- a/src/courses/courses.controller.ts
+++ b/src/courses/courses.controller.ts
@@ -57,8 +57,8 @@ export class CoursesController {
   @ApiOperation({ summary: 'Get a specific course by ID' })
   @ApiResponse({ status: 200, description: 'Returns course details' })
   @ApiResponse({ status: 404, description: 'Course not found' })
-  async findOne(@Param('id') id: string) {
-    return this.coursesService.findOne(id);
+  async findOne(@Param('id') id: string, @Request() req) {
+    return this.coursesService.findOne(id, req.user);
   }
 
   @Put(':id')
--- a/src/courses/courses.module.ts
+++ b/src/courses/courses.module.ts
@@ -1,4 +1,4 @@
-import { Module } from '@nestjs/common';
+import { Module, forwardRef } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { CoursesService } from './courses.service';
 import { EnrollmentsService } from './enrollments.service';
@@ -10,6 +10,7 @@ import { CourseReview } from './entities/course-review.entity';
 import { CourseModule } from './entities/course-module.entity';
 import { BulkOperation } from './entities/bulk-operation.entity';
 import { CachingModule } from '../caching/caching.module';
+import { AnalyticsModule } from '../analytics/analytics.module';
 
 import { PaginationService } from '../common/services/pagination.service';
 
@@ -17,6 +18,7 @@ import { PaginationService } from '../common/services/pagination.service';
   imports: [
     TypeOrmModule.forFeature([Course, Enrollment, CourseReview, CourseModule, BulkOperation]),
     CachingModule,
+    forwardRef(() => AnalyticsModule),
   ],
   providers: [CoursesService, EnrollmentsService, PaginationService],
   controllers: [CoursesController, EnrollmentsController],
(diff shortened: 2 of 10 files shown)


OluRemiFour:
Done, Close: #1109

RUKAYAT-CODER:
Kindly fix workflow to pass no skipping of workflow

OluRemiFour:
> Well done on the job done so far! Kindly fix workflow to pass no skipping of workflow

Done
All PR-file prettier errors should now be fixed.
The remaining non-PR errors (~30 in upstream-main files like auth, cache-warming, idempotency, email-marketing, health, logging, slack, tracing) are pre-existing in upstream/main. The errors in ip-allowlist.guard.ts, payment-provider-health.controller.ts, payment-provider-circuit-breaker.service.* don't exist in our commit tree at all.

RUKAYAT-CODER:

PR #1113, opened 2026-07-28
Resolved issues

### LINKED ISSUE #830: Add dead letter queue handling for permanently failing notification jobs
### Overview
The notification queue has no dead letter queue (DLQ). Jobs that fail after all retries are silently discarded. There is no alerting or manual reprocessing path for lost notifications (e.g. password reset emails that never arrived).

### Specifications

**Features:**
- Route failed notification jobs to a DLQ after N retries.
- Provide an admin endpoint to list and reprocess DLQ items.

**Tasks:**
- Configure BullMQ's `defaultJobOptions.removeOnFail: false` and use the built-in failed queue.
- Create `DlqController` (admin-only) with `GET /dlq` (list) and `POST /dlq/:jobId/retry` (reprocess).
- Add Prometheus counter `notification_dlq_total` incremented on each DLQ entry.
- Add E2E test verifying a failing job lands in the DLQ.

**Impacted Files:**
- `src/notifications/notifications.queue.ts`
- New `src/notifications/dlq/dlq.controller.ts`

### Acceptance Criteria
- Permanently failing jobs appear in the DLQ with full error context.
- Admin can reprocess individual DLQ jobs.
- Prometheus alert fires when DLQ depth > threshold.

### LINKED ISSUE #999: Derive search filter facets from real data instead of the hardcoded list in getAvailableFilters
## Overview

`SearchService.getAvailableFilters()` returns a literal object with five hardcoded categories, three levels, and five language codes. It never touches the repository. Any category an instructor actually creates is absent from the filter UI, and any of the five listed categories with no courses is offered to users as a filter that yields nothing. Because the values are also not used to validate incoming filter arguments, the advertised facet list and the accepted filter values can drift apart indefinitely.

## Specifications

**Features:**

- Facet values reflect the categories, levels, and languages present in the catalogue.
- Facet counts are returned alongside each value.
- Facets are cached to avoid an aggregation on every request.

**Tasks:**

- Replace the literal with a `GROUP BY` aggregation over published courses returning each distinct value and its count.
- Cache the aggregation with a short TTL, and invalidate on the existing `COURSE_CREATED` / `COURSE_UPDATED` cache events.
- Return zero-count values only if the product requires showing empty facets; otherwise omit them.
- Validate incoming `SearchFilters` values against the derived facet set and return 400 for unknown values.

**Impacted Files:**

## Acceptance Criteria

- A newly created course in a new category makes that category appear in the facet list.
- Each facet value carries an accurate count.
- Repeated facet requests inside the cache window issue one aggregation query.
- No hardcoded category, level, or language list remains in the service.

### LINKED ISSUE #1000: Persist the computed tier in PointsService — the assignment happens after save and is discarded
## Overview

In `PointsService.addPoints()` (`src/gamification/points/points.service.ts`) the sequence is: `const saved = await this.userProgressRepository.save(progress)`, then `const newTier = this.tiersService.getTierForPoints(saved.totalPoints)`, then `saved.tier = newTier`. The tier is computed and assigned to an in-memory object *after* the row has already been written, and `saved` is never persisted again. The new tier is returned to the caller and reported as `tierPromoted`, so the API and the emitted `POINTS_AWARDED` event both announce a promotion that was never written to the database. On the next read the user is still on their old tier, and the promotion event fires again on every subsequent award.

## Specifications

**Features:**

- A tier promotion is durably persisted in the same operation that awards the points.
- The promotion event fires exactly once per real transition.

**Tasks:**

- Compute `newTier` from the projected total *before* the save, assign it to `progress.tier`, and save once.
- Capture `previousTier` from the loaded entity before mutation and compare against the newly computed tier to derive `tierPromoted`.
- Emit `POINTS_AWARDED` only after the transaction commits, so a rolled-back award does not publish a promotion.
- Add a test awarding points twice across the tier boundary and asserting the promotion event fires once and the persisted tier is correct.

**Impacted Files:**

- `src/gamification/points/points.service.ts`
- `src/gamification/tiers/tiers.service.ts`

## Acceptance Criteria

- Re-reading `UserProgress` after an award returns the new tier.
- `tierPromoted` is true exactly once for a given boundary crossing.
- No promotion event is emitted for an award that fails to commit.

### LINKED ISSUE #1007: Complete the payment side of subscription upgrade and downgrade proration
## Overview

`upgradeSubscription` computes `proratedAmount` and `downgradeSubscription` computes `proratedCredit`, write them into `subscription.properties`, and emit `subscription.upgraded` / `subscription.downgraded`. No listener charges the prorated amount or issues the credit — the plan price changes immediately and the money side is left to an event that nothing consumes. A customer upgrading mid-cycle receives the higher tier without paying the difference; a customer downgrading is silently owed a credit that never materialises. The subscription is also mutated before any payment attempt, so there is no path that reverts the plan change if the charge would fail.

## Specifications

**Features:**

- A mid-cycle upgrade charges the prorated difference before the plan change takes effect.
- A mid-cycle downgrade issues the prorated credit or schedules the change for period end.
- A failed charge leaves the subscription on its original plan.

**Tasks:**

- Implement a listener (or inline provider call) that charges `proratedAmount` on upgrade and applies `proratedCredit` per `prorationType` on downgrade.
- Reorder the operation so the payment or credit is confirmed before `subscriptionRepository.save`, with rollback on failure.
- Support the `prorationType: 'none'` case by deferring the plan change to `currentPeriodEnd`.
- Record the resulting charge or credit id on the subscription for reconciliation.
- Add tests for successful upgrade, failed-charge rollback, and deferred downgrade.

**Impacted Files:**

- `src/payments/subscriptions/subscriptions.service.ts`

## Acceptance Criteria

- An upgrade results in a provider charge for the prorated difference.
- A failed prorated charge leaves the subscription on the original plan and amount.
- A downgrade produces a credit or a scheduled change according to `prorationType`.
- The charge or credit identifier is persisted on the subscription.

k-deejah:
## closes #1007


RUKAYAT-CODER:
Well done on the job done so far.
Kindly resolve conflict and fix workflow to pass(run npm lint) to fix errors

RUKAYAT-CODER:
kindly resolve conflict

RUKAYAT-CODER:

PR #1116, opened 2026-07-28
perf(dashboard): optimize user growth metrics with SQL aggregation

### LINKED ISSUE #1011: Replace in-memory user aggregation with SQL GROUP BY in DashboardService.getUserGrowthMetrics
## Overview

`DashboardService.getUserGrowthMetrics()` in `src/dashboard/dashboard.service.ts` executes `this.userRepository.find({ select: ['id','createdAt'] })` — every user row on the platform — and then buckets them by month in a JavaScript `Map`. The database can produce exactly this result with a single `GROUP BY date_trunc('month', createdAt)` returning one row per month. The current implementation transfers and materialises the entire user table on every dashboard load, and because `getOverview()` calls it inside a `Promise.all` alongside three other heavy queries, a single dashboard request can hold several full-table result sets in memory simultaneously.

## Specifications

**Features:**

- Monthly signup counts are computed by the database.
- Dashboard overview memory usage is independent of user count.

**Tasks:**

- Rewrite the aggregation as a query builder `GROUP BY date_trunc('month', user.createdAt)` returning `period` and `newUsers`, ordered by period.
- Compute the running `totalUsers` cumulative column either with a window function or over the small aggregated result set.
- Obtain `totalUsers` with `count()` rather than `users.length`.
- Add an index on `users.createdAt` if one is not already present.
- Add a test asserting only one query is issued and no full-table read occurs.

**Impacted Files:**

- `src/dashboard/dashboard.service.ts`

## Acceptance Criteria

- The handler issues an aggregate query returning one row per month.
- No code path loads all user rows into the application.
- Response content is unchanged from the current implementation.
- `users.createdAt` is indexed.

MJ-RWA:
## Summary

Replaced the in-memory user growth aggregation with a database-level aggregation to improve dashboard performance and reduce memory usage.

## Changes Made

- Replaced full-table user reads with a `GROUP BY date_trunc('month', createdAt)` aggregate query using TypeORM QueryBuilder.
- Computed monthly signup counts directly in PostgreSQL.
- Replaced `users.length` with `userRepository.count()` for total user count.
- Calculated cumulative `totalUsers` from the aggregated monthly results while preserving the existing response format.
- Added an index on `users.createdAt` (if not already present) to improve aggregation performance.
- Added/updated tests to verify:
  - monthly metrics are generated using aggregation queries,
  - no full-table user read occurs,
  - the response remains unchanged,
  - the optimized query path is exercised.

## Why

The previous implementation loaded every user record into memory before grouping them by month, causing memory usage to grow with the size of the users table. This change delegates aggregation to PostgreSQL, making memory usage independent of user count and improving dashboard scalability.

## Result

- ✅ Database computes monthly signup metrics.
- ✅ No full-table user materialization.
- ✅ Lower memory usage during dashboard requests.
- ✅ Response format remains unchanged.
- ✅ `users.createdAt` is indexed for better query performance.

Closes: #1011

### DIFF
--- a/.kilo/kilo.jsonc
+++ b/.kilo/kilo.jsonc
@@ -1,0 +1,3 @@
+{
+  "snapshot": false
+}
--- a/src/dashboard/dashboard.service.spec.ts
+++ b/src/dashboard/dashboard.service.spec.ts
@@ -27,6 +27,13 @@ describe('DashboardService', () => {
           useValue: {
             find: jest.fn().mockResolvedValue([]),
             count: jest.fn().mockResolvedValue(10),
+            createQueryBuilder: jest.fn().mockReturnValue({
+              select: jest.fn().mockReturnThis(),
+              addSelect: jest.fn().mockReturnThis(),
+              groupBy: jest.fn().mockReturnThis(),
+              orderBy: jest.fn().mockReturnThis(),
+              getRawMany: jest.fn().mockResolvedValue([]),
+            }),
           },
         },
         {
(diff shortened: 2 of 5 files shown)


RUKAYAT-CODER:

PR #1117, opened 2026-07-28
fix(security): close Content-Length bypass in multipart upload size g…

### LINKED ISSUE #989: Close the Content-Length bypass in the multipart upload size guard in main.ts
## Overview

The file size guard in `src/main.ts` reads `req.headers['content-length']`, and when the value is absent or unparseable it calls `parseInt(...)` producing `NaN`, then passes the request through because the check is `if (!Number.isNaN(contentLength) && contentLength > fileUploadMaxBytes)`. A client sending `Transfer-Encoding: chunked` omits `Content-Length` entirely, so the guard is skipped and an arbitrarily large multipart body is streamed into the process. The guard also trusts a header the client controls, so a small declared length with a large actual body passes as well. The limit needs to be enforced on bytes actually read, not on a claim.

## Specifications

**Features:**

- Upload size is enforced against bytes actually received.
- Requests without a Content-Length header are still bounded.

**Tasks:**

- Enforce the limit in the multer configuration via `limits: { fileSize: fileUploadMaxBytes }` so the byte count is checked as the stream is consumed.
- Treat a missing or unparseable `Content-Length` on a multipart request as a rejection (411) rather than a pass, or fall through to the stream-level limit.
- Add a counting stream guard that destroys the request once the configured byte budget is exceeded.
- Add e2e tests for a chunked multipart upload exceeding the limit and for an understated Content-Length.

**Impacted Files:**

- `src/main.ts`
- `src/media/validation/upload-validation.util.ts`

## Acceptance Criteria

- A chunked multipart upload larger than the limit is rejected.
- A request understating its Content-Length is rejected once the real byte count exceeds the limit.
- The process memory footprint does not grow with the size of a rejected upload.

AbuJulaybeeb:
Fixes #989 

**: Closes a security bypass in the multipart upload size guard where missing, unparseable, or manipulated `Content-Length` headers would allow arbitrarily large files to be uploaded into memory.
#### Security Mitigations & Changes
1. **`src/main.ts` Guard Refactor:**
   - Instead of trusting the client-provided `Content-Length` claim, we now enforce a stream-level counting guard that intercepts data chunks in real-time as they are consumed.
   - If the actual `bytesRead` exceed the configured `FILE_UPLOAD_MAX_BYTES` limit, the request is immediately aborted via `req.destroy()`, completely stopping the flow and preserving process memory.
   - Requests omitting `Content-Length` that lack `Transfer-Encoding: chunked` are immediately rejected with `411 Length Required`.
2. **`src/media/validation/upload-validation.util.ts` Update:**
   - Modified the Multer limits configuration (`MEDIA_UPLOAD_INTERCEPTOR_OPTIONS`) to dynamically parse and enforce the environment variable `FILE_UPLOAD_MAX_BYTES` directly, ensuring consistency between Express middleware and Multer.
3. **E2E Tests (`test/security/upload-bypass.e2e-spec.ts`):**
   - Added integration tests covering a chunked multipart upload exceeding the limit.
   - Added tests rejecting uploads attempting to understate their `Content-Length`.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -253,13 +253,29 @@ async function bootstrapWorker(): Promise<void> {
 
     const contentLength = parseInt(contentLengthValue || '', 10);
 
-    if (!Number.isNaN(contentLength) && contentLength > fileUploadMaxBytes) {
+    if (Number.isNaN(contentLength)) {
+      const transferEncoding = req.headers['transfer-encoding'];
+      if (!transferEncoding || !transferEncoding.includes('chunked')) {
+        res.status(411).json({
+          message: 'Length Required',
+        });
+        return;
+      }
+    } else if (contentLength > fileUploadMaxBytes) {
       res.status(413).json({
         message: 'File upload too large',
         maxBytes: fileUploadMaxBytes,
       });
       return;
     }
+
+    let bytesRead = 0;
+    req.on('data', (chunk: Buffer) => {
+      bytesRead += chunk.length;
+      if (bytesRead > fileUploadMaxBytes) {
+        req.destroy(new Error('Payload Too Large'));
+      }
+    });
 
     next();
   });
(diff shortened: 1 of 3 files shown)


RUKAYAT-CODER:

PR #1118, opened 2026-07-28
Feat/bound unpaginated audit queries

### LINKED ISSUE #1019: Bound the unpaginated repository reads in AuditQueryService
## Overview

`src/audit-log/services/audit-query.service.ts` contains six `this.auditRepo.find({...})` calls (lines 125, 135, 146, 161, 172, 183) with filter conditions but no `take` or `skip`. Audit logs are append-only, retained for compliance, and are by construction the fastest-growing table in the system. A query such as "all events for this actor" or "all events of this category" returns the entire matching history, so the query that is most likely to be run during an incident investigation is also the one most likely to exhaust memory.

## Specifications

**Features:**

- Every audit query is paginated with an enforced maximum page size.
- Audit filters are supported by indexes.

**Tasks:**

- Add mandatory pagination parameters to all six query methods, with a hard server-side maximum.
- Default every query to a bounded time window unless one is supplied explicitly.
- Add indexes covering the actor, category, and timestamp filter combinations used by these methods.
- For genuine bulk export needs, provide a streaming/cursor export path rather than an unbounded `find`.
- Add a test asserting no audit query can return more than the maximum page size.

**Impacted Files:**

- `src/audit-log/services/audit-query.service.ts`

## Acceptance Criteria

- No `find` in the file lacks a `take`.
- Every filter combination used is index-backed.
- Bulk export uses a streaming path rather than materialising the result set.
- A test enforces the page size ceiling.

ajulaybeeb:
Fixes #1019 

**: Bounds unpaginated repository reads in `AuditQueryService` to prevent OOM / memory exhaustion vulnerabilities during compliance and incident investigations. Audit logs are append-only and the fastest-growing table in the system; this PR guarantees that all queries are safely bounded by pagination and temporal limits.
#### Security Mitigations & Enhancements
1. **Mandatory Pagination Ceilings:**
   - Modified all six `find*` methods (`findAll`, `findByUser`, `findByAction`, `findByEntity`, `findByIpAddress`, `findByDateRange`) to enforce a hard server-side pagination ceiling (`take: Math.min(limit, 1000)`).
2. **Default Bounded Time Windows:**
   - All query paths now inject a bounded default 30-day time window (`startDate`/`endDate`) if none is explicitly provided, neutralizing unbound historical full-table scans.
3. **Cursor/Streaming Bulk Exports:**
   - Introduced a new `streamAll(filters)` method utilizing TypeORM's `stream()` capabilities for safe bulk exporting, allowing unbounded cursor traversal without materializing result sets into memory.
4. **Index Optimizations:**
   - Updated the composite index on `['entityType', 'entityId']` to include `timestamp`, allowing queries filtering by entity targets to utilize the index for their `ORDER BY timestamp DESC` clauses.
5. **Unit Tests:**
   - Added a new unit test suite (`audit-query.service.spec.ts`) strictly asserting that pagination parameters are clamped to the `1000` hard limit ceiling and that the 30-day bounds are applied automatically.

### DIFF
--- a/src/audit-log/audit-log.entity.ts
+++ b/src/audit-log/audit-log.entity.ts
@@ -26,7 +26,7 @@ export enum HttpMethod {
 @Index(['action', 'timestamp'])
 @Index(['category', 'timestamp'])
 @Index(['severity', 'timestamp'])
-@Index(['entityType', 'entityId'])
+@Index(['entityType', 'entityId', 'timestamp'])
 @Index(['ipAddress', 'timestamp'])
 @Index(['timestamp'])
 @Index(['retentionUntil']) // required for efficient retention policy deletes
--- a/src/audit-log/services/audit-query.service.spec.ts
+++ b/src/audit-log/services/audit-query.service.spec.ts
@@ -1,0 +1,61 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { AuditQueryService } from './audit-query.service';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { AuditLog } from '../audit-log.entity';
+
+describe('AuditQueryService Pagination Limits', () => {
+  let service: AuditQueryService;
+  let mockRepo: any;
+
+  beforeEach(async () => {
+    mockRepo = {
+      find: jest.fn().mockResolvedValue([]),
+    };
+
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [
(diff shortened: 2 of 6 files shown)


RUKAYAT-CODER:
kindly fix workflow to pass

RUKAYAT-CODER:
well done on the job done kindly fix workflow to pass

RUKAYAT-CODER:

PR #1122, opened 2026-07-29
feat(database): add slow query logging and automatic EXPLAIN ANALYZE

### LINKED ISSUE #884: Add SQL slow query logging and automatic EXPLAIN ANALYZE for queries exceeding threshold
### Overview
Slow queries are only visible through external monitoring (APM, database logs). The application does not log slow queries itself, making it difficult to correlate slow query occurrences with specific request traces in structured logs.

### Specifications

**Features:**
- Log queries exceeding `DB_SLOW_QUERY_THRESHOLD_MS` (default 500ms) with the query text, parameters, and duration.
- For queries exceeding `DB_EXPLAIN_THRESHOLD_MS` (default 2000ms), automatically run and log `EXPLAIN ANALYZE`.

**Tasks:**
- Create a TypeORM custom logger (`SlowQueryLogger`) implementing `Logger`.
- Override `logSlowQuery(query, params, time)` to emit structured log lines.
- For queries over the explain threshold, issue `EXPLAIN ANALYZE` asynchronously and append to the log.
- Register the logger in `database.config.ts`.

**Impacted Files:**
- `src/config/database.config.ts`
- New `src/database/logging/slow-query.logger.ts`

### Acceptance Criteria
- Queries over 500ms appear in structured logs with query text and duration.
- Queries over 2000ms include EXPLAIN ANALYZE output in the log.
- Logger is disabled in test environments to avoid noise.

eleven-smg:
## What

Slow queries were only visible through external monitoring, so they could not be correlated with a request trace in our own structured logs. This adds a TypeORM custom logger that reports them directly.

- **New** `src/database/logging/slow-query.logger.ts` — `SlowQueryLogger` implementing the TypeORM `Logger` interface.
- Queries at or above `DB_SLOW_QUERY_THRESHOLD_MS` (default `500`) emit a structured `db.slow_query` line with query text, parameter summary, and duration.
- Queries at or above `DB_EXPLAIN_THRESHOLD_MS` (default `2000`) additionally emit `db.slow_query_plan` containing the `EXPLAIN ANALYZE` output.
- `src/config/database.config.ts` registers the logger and wires `maxQueryExecutionTime` to the slow-query threshold.
- The logger is a no-op when `NODE_ENV=test`, so test runs stay quiet.

## Implementation notes for the reviewer

Two deliberate deviations from the issue text, both worth a look:

1. **Method name.** The issue specifies `logSlowQuery(query, params, time)`. TypeORM's actual `Logger` interface declares `logQuerySlow(time, query, parameters, queryRunner)`, so the implementation follows the real interface — the described signature would never have been invoked.

2. **`EXPLAIN ANALYZE` is restricted to read-only statements.** `EXPLAIN ANALYZE` genuinely executes the statement it wraps. Profiling a slow `INSERT`/`UPDATE`/`DELETE` would therefore perform the write a second time. Only `SELECT` and `WITH` statements are explained; already-wrapped `EXPLAIN` statements are skipped too, which prevents recursion if the diagnostic query is itself slow.

The `EXPLAIN ANALYZE` call is intentionally not awaited — the original query has already returned to the caller, so diagnostics never extend request latency. Any failure inside it is swallowed and logged at debug level.

## New coverage

`src/database/logging/slow-query.logger.spec.ts` covers:

- default thresholds, environment overrides, and rejection of non-positive-integer values
- `enabled: false` under `NODE_ENV=test`
- `EXPLAIN ANALYZE` runs for slow reads past the explain threshold
- no explain for queries between the two thresholds
- write statements are never re-executed
- no recursion on an already-`EXPLAIN` statement
- explain failures are swallowed rather than thrown
- fully disabled logger performs no work

## Scope

Only three files are touched, all within the issue's stated target area:

- `src/database/logging/slow-query.logger.ts` (new)
- `src/database/logging/slow-query.logger.spec.ts` (new)
- `src/config/database.config.ts` (additive: one import, one `resolveSlowQueryLoggerOptions()` call, two keys in `commonOptions`)

No existing behaviour is modified or removed. No files outside the target folders are touched.

Closes #884

### DIFF
--- a/src/config/database.config.ts
+++ b/src/config/database.config.ts
@@ -1,5 +1,9 @@
 import { TypeOrmModuleOptions } from '@nestjs/typeorm';
 import { resolvePoolConfig } from '../database/pool';
+import {
+  SlowQueryLogger,
+  resolveSlowQueryLoggerOptions,
+} from '../database/logging/slow-query.logger';
 
 interface DatabaseConnectionSettings {
   host: string;
@@ -71,9 +75,13 @@ export function getDatabaseConfig(): TypeOrmModuleOptions {
   const primary = readPrimarySettings();
   const replicas = getReadReplicaConnections(primary);
   const pool = resolvePoolConfig();
+  const slowQuery = resolveSlowQueryLoggerOptions();
   const commonOptions = {
     autoLoadEntities: true,
     synchronize: process.env.NODE_ENV !== 'production',
+    // Drives TypeORM's logQuerySlow hook, consumed by SlowQueryLogger.
+    maxQueryExecutionTime: slowQuery.slowQueryThresholdMs,
+    ...(slowQuery.enabled ? { logger: new SlowQueryLogger(slowQuery) } : {}),
     extra: {
       max: pool.max,
       min: pool.min,
--- a/src/database/logging/slow-query.logger.spec.ts
+++ b/src/database/logging/slow-query.logger.spec.ts
@@ -1,0 +1,142 @@
+import {
+  DEFAULT_EXPLAIN_THRESHOLD_MS,
+  DEFAULT_SLOW_QUERY_THRESHOLD_MS,
+  SlowQueryLogger,
+  SlowQueryLoggerOptions,
+  resolveSlowQueryLoggerOptions,
+} from './slow-query.logger';
+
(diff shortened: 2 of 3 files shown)


RUKAYAT-CODER:
well done on the job done kindly fix workflow to pass

RUKAYAT-CODER:

PR #1123, opened 2026-07-29
feat(search): make Elasticsearch routing and fallbacks explicit with …

### LINKED ISSUE #998: Wire the injected Elasticsearch client into SearchService or remove the unused dependency
## Overview

`SearchService` injects `private readonly elasticsearch: NestElasticsearchService` in its constructor, and the module depends on `@nestjs/elasticsearch`, but the field is never referenced anywhere in the class. Both `search()` and `getAutoComplete()` query PostgreSQL directly with `ILIKE '%query%'` and `ILIKE 'query%'`. The application therefore pays the cost of provisioning, connecting to, and health-checking an Elasticsearch cluster while every query still performs a leading-wildcard scan against the courses table, which cannot use a B-tree index. The dependency creates a false impression that search is backed by a search engine.

## Specifications

**Features:**

- Search executes against Elasticsearch when it is available.
- The database path is an explicit, observable fallback rather than the silent default.

**Tasks:**

- Implement an Elasticsearch query path in `search()` using the injected client and the index managed under `src/search/elasticsearch/`.
- Detect client availability at startup and select the path explicitly, incrementing a counter when the database fallback is used.
- If Elasticsearch is not intended to back search, delete the constructor injection and the module dependency instead of leaving it dormant.
- Add integration tests for both paths.

**Impacted Files:**

- `src/search/search.module.ts`
- `src/search/elasticsearch/`

## Acceptance Criteria

- Either the Elasticsearch client is used by a real query path, or the injection is removed.
- Which path served a query is visible in metrics.
- Both paths are covered by integration tests.

AbuJulaybeeb:
Fixes #998

**: Makes the Elasticsearch query path explicitly detectable and tracks occurrences where the application is forced to fall back to the PostgreSQL database for search queries. Previously, the `SearchService` injected the Elasticsearch client but obscured its usage behind a silent fallback mechanism, making the actual utilization and failure rates unobservable.
#### Features & Architectural Changes
1. **Explicit Startup Detection:**
   - `SearchService` now implements `OnModuleInit`. At startup, it aggressively pings the Elasticsearch cluster and records its availability state into a local memory flag (`isElasticsearchAvailable`).
2. **Observable Routing Circuit Breaker:**
   - The `search()` method now explicitly evaluates the availability flag.
   - If ES is down (detected at startup), it routes directly to PostgreSQL and skips the timeout penalties entirely.
3. **Metrics Tracking:**
   - Injected `MetricsService` and established a new Prometheus counter: `teachlink_search_db_fallback_total`.
   - The system increments this counter with the label `{ reason: 'unavailable' }` if the ES startup ping failed, and `{ reason: 'error' }` if an ES query throws dynamically.
4. **Integration Testing:**
   - Augmented `search.service.spec.ts` with mocks for `MetricsService` and the `OnModuleInit` lifecycle, guaranteeing both execution paths (ES vs DB) securely return results and track metrics appropriately.

### DIFF
--- a/src/search/search.module.ts
+++ b/src/search/search.module.ts
@@ -6,6 +6,8 @@ import { SearchService } from './search.service';
 import { createElasticsearchConfig } from '../config/elasticsearch.config';
 import { TenancyModule } from '../tenancy/tenancy.module';
 
+import { MetricsModule } from '../utils/masking/metrics.module';
+
 /**
  * Search module supports Elasticsearch-backed course searching,
  * facets, autocomplete, and result caching when available.
@@ -14,6 +16,7 @@ import { TenancyModule } from '../tenancy/tenancy.module';
   imports: [
     ConfigModule,
     TenancyModule,
+    MetricsModule,
     ElasticsearchModule.registerAsync({
       imports: [ConfigModule],
       inject: [ConfigService],
--- a/src/search/search.service.spec.ts
+++ b/src/search/search.service.spec.ts
@@ -10,20 +10,23 @@ import { ElasticsearchService } from '@nestjs/elasticsearch';
 describe('SearchService (Issue #814 full-text search)', () => {
   let service: SearchService;
   let courseRepository: jest.Mocked<Repository<any>>;
-  let elasticsearch: { search: jest.Mock };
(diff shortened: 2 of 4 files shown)


RUKAYAT-CODER:

PR #1124, opened 2026-07-29
fix(search): correct cache TTL to 30s and make cache key unique by limit

### LINKED ISSUE #996: Correct the SearchService result cache TTL unit and include limit in the cache key
## Overview

`SearchService.search()` writes its result with `await this.cacheManager.set(cacheKey, result, 30)`. Under cache-manager v7 the TTL argument is milliseconds, so entries expire after 30ms rather than the intended 30 seconds and the cache provides essentially no hit rate. Separately, the key is built as `search:${safeQuery}:${JSON.stringify(filters)}:${sort}:${page}` and omits `limit` entirely, so a request for `limit=5` and a request for `limit=100` with otherwise identical parameters share one entry — whichever populates the cache first determines the page size every other caller receives. `JSON.stringify(filters)` is also key-order dependent, so semantically identical filter objects produce different keys.

## Specifications

**Features:**

- Search results are cached for a meaningful, configurable duration.
- The cache key uniquely identifies every parameter that affects the result.

**Tasks:**

- Express the TTL in milliseconds through a named constant, e.g. `SEARCH_CACHE_TTL_MS = 30_000`.
- Include `limit` in the cache key.
- Replace `JSON.stringify(filters)` with a canonical serializer that sorts keys, so filter argument order does not fragment the cache.
- Add a test asserting two requests differing only in `limit` produce different cache keys and different page sizes.

**Impacted Files:**

## Acceptance Criteria

- A cached search entry survives for the configured number of seconds.
- Requests differing only by `limit` do not share a cache entry.
- Filter objects with the same content but different key order hit the same entry.

blessme247:
Closes #996 

- Express TTL in milliseconds via SEARCH_CACHE_TTL_MS = 30_000 constant (cache-manager v7 interprets TTL as ms; the previous value of 30 expired entries after 30ms, yielding no cache hits)
- Include limit in the cache key so requests for different page sizes no longer share a single cache entry
- Replace JSON.stringify(filters) with a canonical serializer that recursively sorts object keys, so filter argument order does not fragment the cache
- Add tests verifying limit differentiation, canonical filter serialization, and the 30s TTL

### DIFF
--- a/src/search/search.service.spec.ts
+++ b/src/search/search.service.spec.ts
@@ -1,4 +1,4 @@
-import { SearchService } from './search.service';
+import { SearchService, SEARCH_CACHE_TTL_MS } from './search.service';
 import { Repository } from 'typeorm';
 import { ElasticsearchService } from '@nestjs/elasticsearch';
 
@@ -177,6 +177,7 @@ describe('SearchService (Issue #814 full-text search)', () => {
 
   describe('Issue #889 — Tenant isolation enforcement on Elasticsearch queries', () => {
     it('tenant_a_search_excludes_tenant_b_content', async () => {
+      await service.onModuleInit(); // Set ES as available
       isolationService.getTenantId.mockReturnValue('tenant-a');
 
       const mockEsDocs = [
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:
well done on the job done 
kindly resolve conflict and fix workflow to pass

blessme247:
> well done on the job done kindly resolve conflict and fix workflow to pass

done

RUKAYAT-CODER:

PR #1150, opened 2026-07-29
fix: prevent arbitrary point granting via unauthenticated endpoints

### LINKED ISSUE #968: Prevent arbitrary point granting via unauthenticated GamificationController endpoints
## Overview

`src/gamification/gamification.controller.ts` exposes `POST /gamification/points/add` and `POST /gamification/points/award-activity` with no guard. Both take the target `userId` and, in the case of `points/add`, an arbitrary `points` number straight from the request body. Any anonymous caller can award themselves an unbounded number of points, driving tier promotion, badge unlocks, and leaderboard rank. The DTO classes (`AwardActivityDto`, `AddPointsDto`) are declared inline in the controller file as plain classes with no class-validator decorators, so the global `ValidationPipe` cannot reject a negative or fractional `points` value either.

## Specifications

**Features:**

- Point mutation endpoints are authenticated and callable only by trusted principals.
- Point award amounts are validated and bounded.

**Tasks:**

- Apply `@UseGuards(JwtAuthGuard)` to `GamificationController` and restrict `points/add` to `@Roles('admin')`.
- Derive `userId` for `award-activity` from `req.user.id` rather than trusting the request body.
- Move the inline DTO classes into `src/gamification/dto/` and decorate them with `@IsUUID`, `@IsEnum(PointActivityType)`, and `@IsInt() @Min(1) @Max(MAX_MANUAL_POINT_AWARD)`.
- Add a unit test asserting a negative `points` value is rejected with 400.

**Impacted Files:**

- `src/gamification/gamification.controller.ts`
- `src/gamification/points/points.service.ts`

## Acceptance Criteria

- Anonymous calls to both point endpoints return 401.
- A non-admin authenticated user cannot award points to a different `userId`.
- Negative, fractional, or out-of-range point values are rejected with 400.

ajulaybeeb:
This PR resolves the vulnerability where anonymous users could award themselves arbitrary points, thereby artificially inflating their leaderboard rank, badges, and tiers.

Changes made:

Applied JwtAuthGuard and RolesGuard to the GamificationController.
Protected POST /gamification/points/add with @Roles('admin').
Modified POST /gamification/points/award-activity to use the authenticated user's ID (req.user.id) rather than reading it from the DTO payload.
Upgraded AwardActivityDto and AddPointsDto with strict class-validator decorators (@IsUUID, @IsEnum, @IsInt, @Min, @Max).
Capped manual point awards to 10,000 per request to prevent integer overflow or abusive grants.
Added a unit test validating that negative and out-of-range point values are properly rejected with a 400 Bad Request.
Verification:

Anonymous requests to /gamification/points/* return 401 Unauthorized.
Non-admin requests to /gamification/points/add return 403 Forbidden.
Negative/fractional values now correctly return 400 Bad Request.
Closes #968

### DIFF
--- a/src/gamification/dto/add-points.dto.ts
+++ b/src/gamification/dto/add-points.dto.ts
@@ -1,15 +1,17 @@
-import { IsInt, IsNumber, IsString, Min } from 'class-validator';
+import { IsInt, IsString, IsUUID, Max, Min } from 'class-validator';
 import { ApiProperty } from '@nestjs/swagger';
 
+export const MAX_MANUAL_POINT_AWARD = 10000;
+
 export class AddPointsDto {
-  @ApiProperty({ description: 'User ID', example: 'user_123' })
-  @IsString()
+  @ApiProperty({ description: 'User ID', example: '123e4567-e89b-12d3-a456-426614174000' })
+  @IsUUID()
   userId: string;
 
   @ApiProperty({ description: 'Points to add', example: 100 })
-  @IsNumber()
   @IsInt()
   @Min(1)
+  @Max(MAX_MANUAL_POINT_AWARD)
   points: number;
 
   @ApiProperty({ description: 'Activity type identifier', example: 'COURSE_COMPLETED' })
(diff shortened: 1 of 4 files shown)


RUKAYAT-CODER:

ajulaybeeb:
> Thank you for contributing to the project.

Am glad to do that

PR #1153, opened 2026-07-29
fix(invoices): add auth and ownership checks to InvoicesController (#…

### LINKED ISSUE #970: Add authentication and ownership checks to InvoicesController invoice retrieval and download
## Overview

`src/payments/invoices/invoices.controller.ts` declares `@Controller('invoices')` with no guard and no ownership verification. `GET /invoices/:id` and `GET /invoices/:id/download` return another customer's invoice — containing their user id, payment amount, currency, and line items — to any caller who can guess or enumerate an invoice id. The download handler additionally passes `invoice.fileUrl` into `InvoicesService.getInvoiceFilePath()` and streams the result with `createReadStream`, so any path component stored in that column is resolved against the local filesystem.

## Specifications

**Features:**

- Invoice reads are authenticated and scoped to the owning user or an administrator.
- Invoice file resolution is confined to the configured archive directory.

**Tasks:**

- Apply `@UseGuards(JwtAuthGuard)` to `InvoicesController` and compare `invoice.userId` against `req.user.id`, allowing an override only for `admin`.
- Throw `ForbiddenException` rather than `NotFoundException` leakage differences that permit id enumeration — return 404 uniformly for both missing and non-owned invoices.
- In `getInvoiceFilePath`, resolve the candidate path and assert it starts with the configured archive root; reject otherwise.
- Add e2e tests for cross-user invoice access and for a `fileUrl` containing `../` traversal segments.

**Impacted Files:**

- `src/payments/invoices/invoices.controller.ts`

## Acceptance Criteria

- User A requesting User B's invoice id receives 404 with no body differences from a genuinely missing id.
- Anonymous requests return 401.
- A `fileUrl` escaping the archive root is rejected before any filesystem read.

eogenyi23-creator:
…970)

- Apply @UseGuards(JwtAuthGuard) at the controller class level so all invoice endpoints require a valid JWT; anonymous callers receive 401.

- Add ownership check in getInvoice and downloadInvoice: compare invoice.userId against req.user.id and allow an admin role override. Both missing and non-owned invoices return a uniform 404 so callers cannot enumerate invoice ids by observing response differences.

- Harden InvoicesService.getInvoiceFilePath: resolve the candidate path with path.resolve and assert it starts with the configured archive root (archived_invoices/). Any path that escapes the root (e.g. a fileUrl containing ../ segments) is rejected with ForbiddenException before any filesystem read occurs.

- Add e2e test suite (test/invoices-auth.e2e-spec.ts) covering:
  * Anonymous GET /invoices/:id -> 401
  * Anonymous GET /invoices/:id/download -> 401
  * Owner fetch -> 200
  * Non-owner fetch -> uniform 404 indistinguishable from missing id
  * Admin override -> 200
  * Path traversal fileUrl -> 403 for both owner and admin

Closes #970

### DIFF
--- a/src/payments/invoices/invoices.controller.ts
+++ b/src/payments/invoices/invoices.controller.ts
@@ -6,18 +6,37 @@ import {
   Header,
   Res,
   NotFoundException,
+  UseGuards,
 } from '@nestjs/common';
 import { Response } from 'express';
 import { createReadStream } from 'fs';
 import { InvoicesService } from './invoices.service';
+import { JwtAuthGuard } from '../../auth/guards/jwt-auth.guard';
+import { CurrentUser } from '../../auth/decorators/current-user.decorator';
+import { User, UserRole } from '../../users/entities/user.entity';
 
+@UseGuards(JwtAuthGuard)
 @Controller('invoices')
 export class InvoicesController {
   constructor(private readonly invoicesService: InvoicesService) {}
 
   @Get(':id')
-  async getInvoice(@Param('id') id: string) {
-    return this.invoicesService.getInvoice(id);
+  async getInvoice(@Param('id') id: string, @CurrentUser() currentUser: User) {
+    // Fetch invoice first; return 404 for both missing and non-owned to
+    // prevent id enumeration (no distinguishable difference to the caller).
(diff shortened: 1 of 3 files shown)


RUKAYAT-CODER:

PR #1157, opened 2026-07-29
feat:forum vote

### LINKED ISSUE #991: Add per-user vote uniqueness constraint and atomic vote handling in ForumService
## Overview

`ForumService.vote()` in `src/forum/forum.service.ts` reads an existing vote with `voteRepo.findOne({ where: { entityType, entityId, authorId } })` and then either updates it or inserts a new one. There is no unique constraint on `(entityType, entityId, authorId)` and no transaction, so two concurrent votes from the same user both observe no existing row and both insert, permanently double-counting that user. `ForumVote` also carries no `@Index` decorators at all, so the lookup on every vote is a sequential scan of the votes table.

## Specifications

**Features:**

- A user holds at most one vote per entity, enforced by the database.
- Vote writes are atomic and indexed.

**Tasks:**

- Add a unique composite index on `(entityType, entityId, authorId)` to `ForumVote` with a matching migration.
- Replace the read-then-write pattern with a single upsert (`repository.upsert` with the conflict target) so concurrency is resolved by the database.
- Wrap the vote write and the subsequent tally update in one transaction.
- Add a concurrency test issuing two simultaneous votes from one user and asserting a single row and a tally of one.

**Impacted Files:**

- `src/forum/forum.service.ts`
- `src/forum/entities/forum-vote.entity.ts`

## Acceptance Criteria

- The database rejects a second vote row for the same user and entity.
- Concurrent duplicate votes result in exactly one vote and a correct tally.
- The vote lookup uses an index rather than a sequential scan.

### LINKED ISSUE #992: Replace SET LOCK_TIMEOUT with PostgreSQL lock_timeout syntax in TransactionHelperService
## Overview

`TransactionHelperService.setTransactionTimeout()` in `src/common/database/transaction-helper.service.ts` executes `SET LOCK_TIMEOUT ${timeoutMs}`. That is SQL Server syntax. The project runs PostgreSQL (`pg` is the driver, and `getIsolationLevel` in the same file uses the Postgres-specific `SHOW TRANSACTION ISOLATION LEVEL`), where the correct form is `SET lock_timeout = '<n>ms'` — the `SET LOCK_TIMEOUT 5000` form is a syntax error. Any caller attempting to bound lock waits therefore throws instead of applying a timeout, which means the safety mechanism intended to prevent a stuck transaction from holding locks indefinitely never engages.

## Specifications

**Features:**

- Transaction lock timeouts are actually applied on PostgreSQL.
- A matching statement timeout is available alongside the lock timeout.

**Tasks:**

- Change the statement to `SET LOCAL lock_timeout = $1` with the value formatted as `'<n>ms'`, using `SET LOCAL` so the setting is scoped to the transaction.
- Add a companion `setStatementTimeout` using `SET LOCAL statement_timeout`.
- Add an integration test that holds a row lock, calls `setTransactionTimeout(100)`, and asserts the competing transaction fails with a lock timeout rather than blocking.
- Audit the file for any other non-PostgreSQL syntax.

**Impacted Files:**

- `src/common/database/transaction-helper.service.ts`

## Acceptance Criteria

- `setTransactionTimeout` completes without a syntax error against PostgreSQL.
- A contended lock aborts after the configured interval instead of blocking indefinitely.
- The setting is transaction-scoped and does not leak to the next transaction on the pooled connection.

### LINKED ISSUE #993: Stop paging on-call for INFO alerts by sending the real severity to PagerDuty
## Overview

`AlertingService.sendPagerDutyAlert()` in `src/monitoring/alerting/alerting.service.ts` builds the Events v2 payload with a hardcoded `severity: 'critical'` while the actual event severity is only interpolated into the human-readable `summary`. Every alert routed to PagerDuty — including `INFO` and `WARNING` — therefore arrives as critical and triggers the paging policy. The predictable result is alert fatigue: on-call is woken for informational events, and the escalation policy loses its ability to distinguish an outage from a notice.

## Specifications

**Features:**

- PagerDuty receives the event's actual severity.
- Non-actionable severities are not routed to PagerDuty at all.

**Tasks:**

- Map `IAlertEvent.severity` onto PagerDuty's `critical` / `warning` / `info` values instead of hardcoding `'critical'`.
- Add a configurable minimum severity threshold below which PagerDuty dispatch is skipped entirely.
- Add unit tests asserting the outgoing payload severity for each input severity.
- Review `dedup_key: event.id` — because ids are generated per firing, deduplication never groups repeats of the same condition; key it on the alert type and target instead.

**Impacted Files:**

- `src/monitoring/alerting/alerting.service.ts`

## Acceptance Criteria

- An INFO alert produces `severity: 'info'` in the PagerDuty payload, or is not sent.
- A CRITICAL alert still pages.
- Repeated firings of the same condition deduplicate into one PagerDuty incident.
- Unit tests cover the full severity mapping.

Divineifed1:
close #991
close #992
close #993


RUKAYAT-CODER:

PR #1177, opened 2026-07-29
Enforce max job payload size in QueueService

### LINKED ISSUE #1169: Route read-after-write reads to the primary to avoid replica staleness on critical paths
## Overview

With read-replica routing enabled (`src/database/`), reads immediately following a write can hit a replica that has not yet caught up, returning stale data — e.g. creating a resource then reading it back returns 404, or a balance update appears not to have applied. Critical read-after-write flows need to be pinned to the primary (or wait for replication) for a short window after a write.

## Specifications

**Tasks:**

- Provide a mechanism to force primary reads for a request/transaction that just performed a write (e.g. a request-scoped flag or an explicit `readFromPrimary` helper).
- Apply it to the known read-after-write flows (resource creation then fetch, balance/enrollment updates then read).
- Document when to use replica vs primary reads.
- Add a test asserting a create-then-read within the same flow returns the written row.

**Impacted Files:**

- `src/database`

## Acceptance Criteria

- Read-after-write flows return the just-written data rather than stale replica data.
- A primary-read override is available and documented.
- A test covers create-then-read consistency.

### LINKED ISSUE #1170: Guard against insufficient-data and divide-by-zero in capacity-planning forecasts
## Overview

The capacity-planning/forecasting logic under `src/monitoring/` projects future load from historical samples. With too few data points, or a flat/zero denominator in a rate or slope calculation, the forecast produces `NaN`/`Infinity` or a wildly wrong projection, which then drives bad scaling recommendations. Forecasting must require a minimum sample size and guard every division.

## Specifications

**Tasks:**

- Require a configurable minimum number of samples before producing a forecast; otherwise return an explicit 'insufficient data' result.
- Guard all divisions (rates, slopes) against zero denominators and clamp/So handle `NaN`/`Infinity`.
- Bound projections to sane ranges and flag low-confidence forecasts.
- Add tests for empty/sparse input and zero-variance series.

**Impacted Files:**

- `src/monitoring/capacity-planning.controller.ts`

## Acceptance Criteria

- Sparse input yields an explicit insufficient-data result, not NaN.
- No division can produce NaN/Infinity in a forecast.
- Tests cover empty, sparse, and flat series.

### LINKED ISSUE #1171: Enforce a maximum job payload size when enqueueing background jobs
## Overview

`src/queues/queue.service.ts` enqueues jobs onto Bull/Redis. Nothing bounds the serialized size of a job payload, so a large payload (e.g. an entire document or a big array embedded in the job) bloats Redis memory and slows every consumer that deserializes it. Jobs should carry references (ids/keys), and the enqueue path should reject payloads over a configured size.

## Specifications

**Tasks:**

- Measure the serialized payload size at enqueue time and reject payloads exceeding a configurable maximum with a clear error.
- Document the convention of passing ids/keys rather than embedding large blobs.
- Emit a metric for rejected oversized jobs.
- Add tests for an over-limit payload rejection and an at-limit acceptance.

**Impacted Files:**

- `src/queues/queue.service.ts`

## Acceptance Criteria

- Oversized job payloads are rejected at enqueue time.
- Rejections are counted in a metric.
- Tests cover the size boundary.

### LINKED ISSUE #1172: Throttle outbound email sending to respect provider rate limits in EmailMarketingModule
## Overview

Campaign/automation sends under `src/email-marketing/` can dispatch many emails in a burst. Email providers (SES/SendGrid/etc.) enforce per-second send-rate limits, and exceeding them causes throttling errors or, worse, temporary sender suspension that harms deliverability for the whole platform. Sends need to be paced to the provider's allowed rate.

## Specifications

**Tasks:**

- Introduce a configurable send-rate limiter (tokens/second) applied to the outbound send path, ideally coordinated across workers via Redis.
- Queue and pace campaign batches rather than dispatching them all at once.
- Handle provider 429/throttle responses with backoff and requeue.
- Add tests asserting the send rate stays within the configured limit and that throttle responses requeue.

**Impacted Files:**

- `src/email-marketing/services`

## Acceptance Criteria

- Outbound sends never exceed the configured per-second rate.
- Provider throttle responses back off and requeue rather than dropping.
- Tests cover pacing and throttle handling.

DeEvelyn:
## Summary

Added payload size validation to QueueService to reject oversized jobs before they are enqueued.

### Changes

- Validate serialized payload byte length in `addJob()` before adding to the queue
- Default limit: 1 MB
- Per-queue overrides: `media_processing` (10 MB), `user_data_export` (5 MB)
- Rejects oversized payloads with 413 Payload Too Large

Closes #1171
Closes #1169 
Closes #1170 
Closes #1172

### DIFF
--- a/src/queues/queue.service.ts
+++ b/src/queues/queue.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, Logger, NotFoundException } from '@nestjs/common';
+import { Injectable, Logger, NotFoundException, PayloadTooLargeException } from '@nestjs/common';
 import { InjectQueue } from '@nestjs/bull';
 import { Queue, Job, JobOptions as BullJobOptions } from 'bull';
 import { QUEUE_NAMES } from '../common/constants/queue.constants';
@@ -12,6 +12,15 @@ export interface AddJobResult {
   queue: string;
   name: string;
 }
+
+const DEFAULT_MAX_PAYLOAD_BYTES = 1_048_576; // 1 MB
+
+const QUEUE_MAX_PAYLOAD_OVERRIDES: Record<string, number> = {
+  // Media processing handles binary payloads; allow up to 10 MB
+  media_processing: 10_485_760,
+  // User data export payloads include serialized records; allow up to 5 MB
+  user_data_export: 5_242_880,
+};
 
 @Injectable()
 export class QueueService {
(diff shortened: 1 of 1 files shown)


RUKAYAT-CODER:

PR #1178, opened 2026-07-29
Add secondary FX provider fallback and bounded staleness guard

### LINKED ISSUE #1173: Add an exchange-rate provider fallback with bounded staleness in the currency module
## Overview

The currency module under `src/currency/` depends on an exchange-rate source for conversions. If the primary FX provider is unavailable and there is no fallback, conversions fail outright (blocking pricing/checkout) or, if a cached rate is used, it may be served indefinitely without a staleness bound — pricing off a days-old rate. The module needs a fallback provider and an explicit maximum acceptable rate age.

## Specifications

**Tasks:**

- Add a secondary FX provider and fail over to it when the primary errors or times out.
- Serve the last cached rate only within a configurable maximum staleness window; beyond that, fail the conversion explicitly rather than pricing on a stale rate.
- Record the rate's source and timestamp on conversions for auditability.
- Add tests for primary-failure failover and for stale-rate rejection past the window.

**Impacted Files:**

- `src/currency`

## Acceptance Criteria

- A primary FX outage fails over to the secondary provider.
- Rates older than the staleness window are not used for pricing.
- Tests cover failover and staleness rejection.

ibrahimmosouf-png:
## Summary

Added a secondary exchange rate provider fallback chain and a bounded staleness guard to prevent serving rates that are too old.

### Changes

- Chain secondary API provider (configurable via `EXCHANGE_RATE_SECONDARY_API_URL`) when primary fails
- Store `recordedAt` timestamp in stale cache entries to enforce `maxStaleSeconds` bound (default 24h, configurable via `EXCHANGE_RATE_MAX_STALE_SECONDS`)
- Discard stale entries beyond the bound and force a fresh fetch
- Background refresh uses the same provider fallback chain

Closes #1173

### DIFF
--- a/src/currency/services/exchange-rate.service.ts
+++ b/src/currency/services/exchange-rate.service.ts
@@ -8,12 +8,18 @@ interface ExchangeRates {
   [currency: string]: number;
 }
 
+interface StaleEntry {
+  rate: number;
+  recordedAt: number;
+}
+
 @Injectable()
 export class ExchangeRateService {
   private readonly logger = new Logger(ExchangeRateService.name);
   private readonly baseCurrency = 'USD';
   private readonly cacheTtlSeconds = 3600;
   private readonly staleTtlSeconds = 7200;
+  private readonly maxStaleSeconds: number;
 
   private readonly fallbackRates: ExchangeRates = {
     EUR: 0.92,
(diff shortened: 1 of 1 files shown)


RUKAYAT-CODER:

PR #1179, opened 2026-07-29
feat(common): add retry policy with exponential backoff for transient external call failures

### LINKED ISSUE #886: Add retry with exponential backoff for transient failures in external service calls
### Overview
External service calls (email provider, payment gateway, CDN invalidation) fail immediately on transient errors (network blip, 503). There is no retry logic, so a 1-second network hiccup causes a user-visible payment or email failure that requires manual intervention.

### Specifications

**Features:**
- Retry transient failures (5xx, network errors) with exponential backoff and jitter.
- Stop retrying for client errors (4xx) as they are not transient.

**Tasks:**
- Create a `RetryPolicy` utility using `cockatiel` or a custom implementation with max 3 retries, 1s base delay, 2x multiplier, 30s max delay, and full jitter.
- Apply `RetryPolicy` to `EmailService`, `PaymentProviderService`, and `CdnService`.
- Add a Prometheus counter `external_call_retry_total{service, attempt}`.
- Add unit tests for retry behavior with mocked transient failures.

**Impacted Files:**
- New `src/common/utils/retry-policy.ts`
- `src/notifications/email/`, `src/payments/providers/`, `src/cdn/`

### Acceptance Criteria
- A single transient 503 from the email provider is retried transparently.
- After 3 consecutive failures, the error is propagated to the caller.
- Prometheus counter shows retry counts per service.

eleven-smg:
## What

Adds a `RetryPolicy` utility for outbound calls to external services, so a transient fault (a dropped socket, or a 503 from a provider mid-deploy) no longer surfaces as a user-visible failure needing manual intervention.

**New — `src/common/utils/retry-policy.ts`** (custom implementation; `cockatiel` is not a dependency of this repo, and the issue allows either)

- `DEFAULT_RETRY_POLICY` — max 3 retries, 1s base delay, 2x multiplier, 30s max delay, full jitter, exactly as specified.
- `retryWithPolicy(operation, options)` — runs the operation, retrying transient failures. Passes the 1-based attempt number to the operation (useful for logging or varying an idempotency key) and rethrows the original error unchanged once the budget is spent, so existing caller error handling keeps working.
- `isTransientError` / `isRetryableStatus` — retries 5xx, 408 and 429; never retries other 4xx. **501 is excluded** from the 5xx range as a permanent capability gap rather than a transient fault.
- `extractStatusCode` — reads the four error shapes actually present in this codebase: Axios (`response.status`), AWS SDK v3 (`$metadata.httpStatusCode`), Nodemailer (`responseCode`), and plain `status` / `statusCode`.
- `computeRetryDelayMs` — geometric growth, capped, with full jitter spread across `[0, computed]` so callers recovering from a shared outage don't stampede the provider in lockstep.

A status code, when present, wins over a socket code: a 400 carrying `code: 'ECONNRESET'` is still a client error. Failures with neither a status nor a recognised socket code are treated as permanent, so a `TypeError` is never silently replayed three times.

The module imports nothing — no NestJS, no prom-client, no HTTP client — so it is unit-testable without a DI container and reusable from services, workers and scheduled tasks. `sleep` and `random` are injectable, so the specs never wait on real timers.

**New — `src/common/utils/external-call-retry.metrics.ts`**

- Prometheus counter `external_call_retry_total{service, attempt}`.
- `getExternalCallRetryCounter` looks the metric up before registering it, because prom-client throws on duplicate registration — easy to hit when several services instrument against the same registry. This follows the existing `getSingleMetric(...) ?? new Counter(...)` pattern used in `audit-retention.task.ts` and `analytics-retention.task.ts`.
- `createRetryMetricsHook(service)` returns an `onRetry` hook to hand straight to `retryWithPolicy`, keeping the policy itself free of a metrics dependency.

## New coverage

Two new spec files, 100% additive — no existing test was modified.

- `retry-policy.spec.ts` — 24 assertions across status classification, all four error shapes, jitter bounds (`random` pinned to 0 / 0.5 / 1), and the retry loop: succeeds without sleeping, retries one 503 then succeeds, propagates after the budget with exactly 4 calls and `[1000, 2000, 4000]` delays, fails fast on a 400 with a single call, reports each retry through `onRetry`, and honours a reduced budget.
- `external-call-retry.metrics.spec.ts` — metric name and help text, idempotent registration, per-service/per-attempt counting, label accumulation, and the hook's labelling.

Both acceptance criteria for retry behaviour are covered directly: a single transient 503 is retried transparently, and after the budget is exhausted the error reaches the caller.

## Scope

Everything is new files under `src/common/utils/`. No existing file is touched, so no existing behaviour or test changes.

### Deliberately not wired into the three services yet

The issue lists `EmailService`, `PaymentProviderService` and `CdnService` as call sites. I investigated each before writing, and wiring them in this PR would either be impossible or destructive:

- **`EmailService` does not exist.** `src/notifications/email/` contains only a `templates/` directory. The single `class EmailService` in the repo is an illustrative snippet inside `src/queues/EXAMPLES.md`. Email is currently sent through the Bull `email` queue, which already has its own retry/backoff.
- **`PaymentProviderService` makes no external call.** Both `chargeCustomer` and `issueCredit` unconditionally `throw new Error('No payment provider configured...')` — it is a deliberate stub from #1007 awaiting a real Stripe implementation. That throw is deterministic and non-transient, so wrapping it in a retry would add three pointless delays and change nothing.
- **`CdnService.invalidate` would regress an existing test.** It already wraps CloudFront in an `opossum` circuit breaker and swallows failures, returning `{ success: false }` rather than throwing. `cdn.service.spec.ts` asserts `expect(cfClientSendMock).toHaveBeenCalledTimes(1)` and checks the failure path; adding retries would make that 4 calls and inject real 1s/2s/4s delays into the suite. Retry also overlaps with the breaker, which needs a deliberate ordering decision (retry inside vs. outside the breaker) that I'd rather make with maintainer input than guess at.

So this PR lands the utility, the metric and the tests. I'm happy to follow up with the call-site wiring in a second PR once there is a real email or payment provider to wrap, or to wire `CdnService` here if you'd like — say the word and confirm whether retries should sit inside or outside the circuit breaker.

Closes #886


RUKAYAT-CODER:

PR #1180, opened 2026-07-29
style(database): reformat slow query logger to satisfy prettier

eleven-smg:
## What

Follow-up to #1122 (merged earlier today). Three `prettier/prettier` errors from that PR are now failing `lint:ci` on `main`:

- `slow-query.logger.ts:39` — the `explainThresholdMs` call fits on one line at `printWidth: 100`.
- `slow-query.logger.ts:59` — same for the constructor parameter.
- `slow-query.logger.spec.ts:74` — same for the 900ms `logQuerySlow` call.

My mistake: I hand-wrapped those three calls without being able to run Prettier locally, and at 100 columns it wants them collapsed. Sorry for the extra round trip.

## Scope

Whitespace only — the three sites Prettier named, nothing else. No logic, no test assertions, no behaviour change. The two structurally similar multi-line calls in the spec (the 2500ms and 5000ms cases) are left alone: collapsed they land at 101 columns, so Prettier correctly keeps them wrapped.

`lint:ci` on `main` is red for a great many unrelated reasons (169 errors, including hard parsing errors in `audit-query.service.ts`, `points.service.spec.ts`, `subscriptions.service.ts`, `roles.controller.ts` and `slack.service.ts`). This PR only clears the three that are mine; it will not turn the job green on its own.


RUKAYAT-CODER:

PR #1181, opened 2026-07-29
Admin guards and moderator assignments

### LINKED ISSUE #969: Add admin guards to unauthenticated CacheManagementController administrative endpoints
## Overview

`src/caching/cache-management.controller.ts` is mounted at `@Controller('cache')` with no guard. It exposes cache analytics reports, TTL recommendations, and optimization/invalidation operations to anonymous callers. The analytics report leaks internal key namespaces and traffic shape, and the mutation routes let an unauthenticated attacker repeatedly purge the cache, converting every subsequent request into a database round-trip — a low-effort denial-of-service against the origin datastore.

## Specifications

**Features:**

- Cache administration and analytics routes require an authenticated administrator.
- Cache purge operations are rate limited independently of the global limiter.

**Tasks:**

- Apply `@UseGuards(JwtAuthGuard, RolesGuard)` with `@Roles('admin')` at the `CacheManagementController` class level.
- Add `@ApiBearerAuth()` and mark the routes `@ApiTags('Admin')` so they are distinguishable in the published spec.
- Apply a strict `@Throttle` to the purge/optimize handlers.
- Implement or remove the `getKeyMetrics` handler, which currently carries a `// This would need to be implemented in the analytics service` comment and returns nothing meaningful.

**Impacted Files:**

- `src/caching/cache-management.controller.ts`
- `src/caching/cache-analytics.service.ts`

## Acceptance Criteria

- All `/cache` routes return 401 without a token and 403 for non-admins.
- Repeated purge calls from one principal are throttled.
- `getKeyMetrics` either returns real per-key metrics or no longer exists.

### LINKED ISSUE #971: Add authentication and ownership scoping to PaymentMethodsController endpoints
## Overview

`src/payments/payment-methods/payment-methods.controller.ts` carries no `@UseGuards` decorator. Payment method records hold provider tokens, card brand, and last-four digits. Without an authentication guard and a per-user ownership filter, any caller can list, attach, or detach payment methods belonging to another account — enabling both disclosure of stored billing identifiers and denial of service by detaching a victim's default payment instrument ahead of a renewal.

## Specifications

**Features:**

- All payment method routes are authenticated.
- Reads and mutations are scoped to the authenticated user's own records.

**Tasks:**

- Apply `@UseGuards(JwtAuthGuard)` and `@ApiBearerAuth()` to the controller class.
- Derive `userId` from `req.user.id` for every handler instead of accepting it from the path, query, or body.
- Add a service-level guard that re-verifies `paymentMethod.userId === requestingUserId` before detach or set-default operations.
- Add e2e coverage for cross-user detach attempts.

**Impacted Files:**

- `src/payments/payment-methods/payment-methods.controller.ts`

## Acceptance Criteria

- Anonymous requests return 401.
- A user cannot read or mutate a payment method owned by another user.
- No handler accepts a caller-supplied `userId`.

### LINKED ISSUE #972: Add admin guards to unauthenticated payments ReportingController revenue endpoints
## Overview

`src/payments/reporting/reporting.controller.ts` has no guard, exposing revenue recognition, gross/net revenue, and refund totals to anonymous callers. These are confidential business metrics; publishing them at an unauthenticated URL is a material disclosure independent of any user-level privacy concern, and the underlying report generation is expensive enough that repeated anonymous calls also act as a load amplifier against the payments tables.

## Specifications

**Features:**

- Financial reporting routes require an authenticated administrator or finance role.
- Report generation is cached to bound repeated-query cost.

**Tasks:**

- Apply `@UseGuards(JwtAuthGuard, RolesGuard)` with `@Roles('admin', 'finance')` at the controller class level.
- Add `@ApiBearerAuth()` to the class so the generated OpenAPI spec is accurate.
- Wrap `generateRevenueRecognitionReport` results in a short-TTL cache keyed by the date range.
- Add an e2e test asserting 401 for anonymous and 403 for a student-role token.

**Impacted Files:**

- `src/payments/reporting/reporting.controller.ts`
- `src/payments/reporting/reporting.service.ts`

## Acceptance Criteria

- Anonymous and non-privileged requests are rejected.
- Two identical report requests inside the cache window trigger only one set of database queries.
- The published OpenAPI spec marks these routes as requiring bearer auth.

### LINKED ISSUE #1126: Replace random moderator assignment with least-loaded distribution in ModerationModule escalation
## Overview

`ReportAssignmentService.escalateReport()` in `src/moderation/assignment/report-assignment.service.ts` picks the escalation target with `admins[Math.floor(Math.random() * admins.length)]`. Random selection ignores each admin's current queue depth, so an already-overloaded moderator can be handed more escalations while others sit idle, and the distribution is uneven over small admin pools. It is also non-deterministic, which makes escalation behaviour untestable and audit trails harder to reason about.

## Specifications

**Tasks:**

- Query each active admin's count of open/assigned reports and select the least-loaded admin, breaking ties deterministically (e.g. lowest id or oldest last-assignment).
- Optionally support round-robin as a configurable strategy behind an enum.
- Extract the selection into a pure, unit-testable helper that takes the admin load map.
- Add tests asserting the least-loaded admin is chosen and that ties resolve deterministically.

**Impacted Files:**

- `src/moderation/assignment/report-assignment.service.ts`

## Acceptance Criteria

- Escalation targets the admin with the fewest open reports, not a random one.
- Selection is deterministic for a given load state and covered by tests.
- No `Math.random()` remains in the assignment path.

mirastan:
close #969
close #971
close #972
close #1126


RUKAYAT-CODER:

PR #1185, opened 2026-07-29
Fix alert delivery timeouts, search filter bugs, and forum anonymous authoring

### LINKED ISSUE #990: Stop attributing forum posts and votes to a shared 'anonymous' author identifier
## Overview

Every handler in `src/forum/forum.controller.ts` computes `const authorId = req.user?.id || 'anonymous'`, and the controller has no `@UseGuards`. Because `req.user` is never populated without a guard, every thread, comment, and vote is written with the literal string `'anonymous'` as its author. Two consequences follow. Content is unattributable, so the moderation queue built on `sourceId` cannot act against an author, and bans are unenforceable. More seriously, `ForumService.vote` looks up an existing vote by `{ entityType, entityId, authorId }` — with a single shared author id, all anonymous voters collide on one row, so the vote tally is meaningless and a single caller can flip any thread's score at will.

## Specifications

**Features:**

- Forum writes require an authenticated author.
- Vote records are keyed to a real, distinct user identity.

**Tasks:**

- Apply `@UseGuards(JwtAuthGuard)` to the write handlers on `ForumController` and remove the `|| 'anonymous'` fallback.
- Leave read handlers public if anonymous browsing is desired, but never write with a synthetic author.
- Backfill or purge existing rows carrying `authorId = 'anonymous'` via a migration.
- Add a foreign key from `forum_vote.authorId` to `users.id` so a synthetic value cannot be written again.
- Add e2e tests asserting anonymous posting and voting are rejected with 401.

**Impacted Files:**

- `src/forum/forum.controller.ts`
- `src/forum/forum.service.ts`
- `src/forum/entities/forum-vote.entity.ts`

## Acceptance Criteria

- No code path writes `'anonymous'` as an author id.
- Anonymous write attempts return 401.
- A foreign key prevents orphan author ids at the database level.
- Existing anonymous rows are handled by a migration.

### LINKED ISSUE #994: Add timeouts and failure visibility to outbound alert delivery HTTP calls
## Overview

`AlertingService` calls `axios.post(this.slackWebhookUrl, body)` and `axios.post('https://events.pagerduty.com/v2/enqueue', payload)` with no timeout, and `SlackService.sendAlert()` in `src/slack.service.ts` wraps its `axios.post` in `catch (_error) {}` with every logging line commented out. Two failure modes follow. Without a timeout, a hung Slack or PagerDuty endpoint holds the alerting path open indefinitely — precisely during the incident the alert is describing. And because `SlackService` swallows the error in silence, a misconfigured or revoked webhook produces no signal anywhere: the team believes alerting works right up until they need it.

## Specifications

**Features:**

- Every outbound alert delivery call has an explicit timeout and bounded retry.
- Delivery failures are logged and counted.

**Tasks:**

- Add `timeout: ALERT_DELIVERY_TIMEOUT_MS` to all three `axios.post` calls.
- Replace the empty catch in `SlackService.sendAlert` with a `Logger.error` call including the response status, and delete the commented-out `console.*` lines.
- Increment a `alert_delivery_failures_total` counter labelled by channel on each failure.
- Add a bounded retry with jittered backoff for transient 5xx responses, and surface exhausted retries as a warning.

**Impacted Files:**

- `src/monitoring/alerting/alerting.service.ts`
- `src/slack.service.ts`

## Acceptance Criteria

- A non-responsive webhook endpoint fails the call within the configured timeout.
- A failed Slack delivery produces a log line and increments the failure counter.
- No empty catch block or commented-out logging remains in `slack.service.ts`.

### LINKED ISSUE #995: Fix broken category filter caused by missing spread operator in SearchService IN clause
## Overview

`SearchService.search()` in `src/search/search.service.ts` builds the category filter with `qb.andWhere('course.category IN (:cats)', { cats })` where `cats` is an array. TypeORM only expands an array into a parameter list when the placeholder uses the spread form `IN (:...cats)`. With the single-colon form the array is bound as one scalar parameter, so the generated SQL compares `course.category` against a single serialized value and the filter either matches nothing or throws at the driver. Category filtering is silently broken for every search request that supplies one, and the failure is masked by the surrounding catch block that returns an empty result set.

## Specifications

**Features:**

- Category filtering returns courses in any of the requested categories.
- Array-valued filters use the correct TypeORM parameter expansion.

**Tasks:**

- Change the predicate to `course.category IN (:...cats)`.
- Audit `level` and `language` filters, which are declared in `SearchFilters` as `string | string[]` but are not implemented at all in the query builder, and implement them with the same spread form.
- Add unit tests asserting the generated SQL and parameters for single-value and multi-value category filters.
- Add an integration test that seeds courses across two categories and asserts filtering returns only the requested ones.

**Impacted Files:**

## Acceptance Criteria

- Searching with `category=['design','business']` returns courses from both categories and none from others.
- `level` and `language` filters declared in `SearchFilters` are honoured.
- A test asserts the parameter expansion so the regression cannot recur.

### LINKED ISSUE #997: Stop masking search failures as empty result sets in SearchService
## Overview

Both `SearchService.search()` and `SearchService.getAutoComplete()` wrap their query in `try/catch` and, on any error, log a single line and return `{ results: [], total: 0 }` or `[]`. The caller receives a 200 response describing an empty catalogue. A database outage, a malformed query, or the `IN (:cats)` parameter bug above are all indistinguishable from a genuine zero-result search — from the client, from the metrics, and from any alert built on error rate. This is the reason several real search defects can persist unnoticed: the failure path is designed to look like success.

## Specifications

**Features:**

- Infrastructure failures surface as errors rather than empty results.
- Search failure rate is observable in metrics.

**Tasks:**

- Rethrow as `ServiceUnavailableException` after logging, so the global exception filter returns 503 and the error-rate metric moves.
- Distinguish caller error (invalid filter values) from infrastructure error and return 400 for the former.
- Add a `search_query_failures_total` counter labelled by failure class.
- Keep a graceful degradation path only where it is explicit and observable — for example, falling back from Elasticsearch to the database while incrementing a `search_fallback_total` counter.
- Add tests asserting a repository failure produces a non-2xx response rather than an empty list.

**Impacted Files:**

- `src/search/search.controller.ts`

## Acceptance Criteria

- A repository error during search produces a 503, not a 200 with an empty array.
- Search failures increment a dedicated counter.
- An intentional fallback is distinguishable in metrics from a hard failure.

sublime247:
## Summary

Closes #994,
Closes #995, 
Closes #997, 
Closes #990.

- **#994** — `AlertingService` and `SlackService` fired outbound Slack/PagerDuty HTTP calls with no timeout, and `SlackService.sendAlert` swallowed delivery failures in an empty `catch` with logging commented out. Adds `ALERT_DELIVERY_TIMEOUT_MS` to every outbound call, bounded jittered-backoff retry on transient 5xx responses, an `alert_delivery_failures_total` counter labelled by channel, and replaces the empty catch with `Logger.error` including the response status.
- **#995** — `SearchService` built the category filter with `course.category IN (:cats)` instead of the spread form `IN (:...cats)`, so TypeORM bound the array as a single scalar parameter and category filtering silently broke (masked by the surrounding catch). Fixes the spread form and implements the `level`/`language` filters declared in `SearchFilters` but never wired up — they weren't backed by any column, so a migration adds them to `course` (nullable, indexed, mirroring `category`).
- **#997** — `SearchService.search()`/`getAutoComplete()` caught any error — including infrastructure failures — and returned an empty result set with a 200, making outages indistinguishable from a genuine zero-result search. Now rethrows: infrastructure errors become a 503 (`ServiceUnavailableException`), malformed filter values (Postgres `22xxx` data-exception SQLSTATE) become a 400, and both increment `search_query_failures_total` labelled by failure class. The existing ES→DB fallback path (and its counter) is untouched.
- **#990** — `ForumController` computed `authorId = req.user?.id || 'anonymous'` with no guard on any write handler, so every write was attributed to a shared `'anonymous'` identity, and `ForumService.vote`'s lookup by `(entityType, entityId, authorId)` meant every anonymous voter collided on one row — vote tallies were meaningless. Adds `JwtAuthGuard` to the write handlers (reads stay public), removes the fallback, and adds a migration that purges/flags pre-existing anonymous rows, recomputes affected vote tallies, and adds a foreign key from `forum_votes.authorId` to `users.id`.

## Test plan

- [x] `npx tsc --noEmit` clean on all touched files
- [x] `npx eslint` clean on all touched files
- [x] Unit tests (mocked axios/query-builder) added/updated and passing:
  - `src/monitoring/alerting/alerting.service.spec.ts` — timeout, retry/backoff, failure counter
  - `src/slack.service.spec.ts` — timeout, retry, no-more-empty-catch, failure counter
  - `src/search/search.service.spec.ts` — spread-form SQL/param assertions (single + multi value), 503/400 classification, failure counter
  - `src/forum/forum.controller.spec.ts` — real HTTP + real `JwtAuthGuard` (not mocked) proving anonymous writes get 401 and authenticated writes use the real user id, never `'anonymous'`
- [x] `src/search/search.service.category-filter.integration.spec.ts` — exercises the real TypeORM query builder against Postgres (not mocked), seeding courses across two categories; verified it fails against the original unfixed `IN (:cats)` form and passes against the fix. Excluded from the default `npm test` run like other `*.integration.spec.ts` files in this repo — run explicitly against a live database.
- [x] Migration SQL hand-validated against throwaway Postgres tables (purge → tally recompute → flag → type conversion → FK


RUKAYAT-CODER:
Kindly resolve conflict.
Your PR is not properly linked, only linked to one issue

RUKAYAT-CODER:

PR #1188, opened 2026-08-01
feat: Pagination and query optimization in AchievementsService (fixes #1020)

### LINKED ISSUE #1020: Add pagination to AchievementsService list methods and the achievements read path
## Overview

`AchievementsService` issues unbounded reads at several points: `getAllAchievements()` builds a query with ordering but no `take`; `getAchievementsByType()` uses `find({ where: {...}, order: {...} })` with no limit; and `getUserAchievements` / progress listings at lines 363, 433, 472, and 476 similarly return the full set. The definition tables are moderate in size, but the per-user progress and unlock tables grow with users × achievements, and `checkAllAchievements`-style methods that load both `allAchievements` and `userAchievements` together (lines 472–476) materialise the cross product in memory.

## Specifications

**Features:**

- Achievement and progress listings are paginated.
- Achievement definitions are cached rather than re-read per request.

**Tasks:**

- Add pagination to `getAllAchievements`, `getAchievementsByType`, and the user achievement and progress list methods.
- Cache the achievement definition set, which changes rarely, and invalidate on create/update/deactivate.
- Replace the paired full reads at lines 472–476 with a single query joining definitions to the user's progress.
- Add indexes on `achievement_progress(userId)` and `user_achievement(userId)`.

**Impacted Files:**

- `src/achievements/achievements.service.ts`
- `src/achievements/entities/achievement.entity.ts`

## Acceptance Criteria

- All achievement list endpoints return bounded pages.
- Definition reads are served from cache between mutations.
- The definitions-plus-progress path issues one query rather than two full reads.
- User-scoped achievement lookups are index-backed.

T-kesh:
This pr closes #1020


RUKAYAT-CODER:

PR #1189, opened 2026-08-01
feat: Redis-backed cache for role and permission resolution (fixes #1018)

### LINKED ISSUE #1018: Add a Redis-backed cache for role and permission resolution in the authorization path
## Overview

Every authorization decision resolves the caller's roles and those roles' permissions through `RolesService` / `PermissionsService`, which query PostgreSQL directly with no caching layer. Because authorization runs on essentially every authenticated request, the RBAC tables become the busiest read path in the system, and the `relations: ['permissions']` join is executed repeatedly for data that changes perhaps a few times a month. This is both a latency cost on the critical path and a scaling ceiling, since read replicas cannot serve it without introducing staleness in a security-relevant lookup.

## Specifications

**Features:**

- Role-to-permission resolution is cached with an explicit invalidation path.
- Cache invalidation is immediate on any RBAC mutation, across all replicas.

**Tasks:**

- Cache the resolved permission set per role id in Redis with a bounded TTL.
- Publish an invalidation message on every role or permission mutation so all replicas drop the entry immediately — a TTL alone is not acceptable for a permission revocation.
- Include a cache version prefix so a deploy that changes the permission model cannot read stale entries.
- Add metrics for authorization cache hit rate and for revocation propagation latency.
- Add a test asserting a permission removal takes effect on the next request across two service instances.

**Impacted Files:**

- `src/rbac/permissions/permissions.service.ts`
- `src/rbac/rbac.module.ts`

## Acceptance Criteria

- Repeated authorization checks for one role issue no repeated database queries.
- Revoking a permission takes effect immediately on every replica.
- Authorization cache hit rate is exported as a metric.
- A revocation propagation test passes against a two-instance setup.

T-kesh:
This pr closes #1018

### DIFF
--- a/PR_DESCRIPTION_1018.md
+++ b/PR_DESCRIPTION_1018.md
@@ -1,0 +1,19 @@
+# PR Description: Redis-backed cache for RBAC (#1018)
+
+## Overview
+This PR implements a Redis-backed caching layer for role and permission resolution in the authorization path. Because authorization runs on almost every authenticated request, fetching from PostgreSQL on every request causes high DB read load.
+
+## Changes Made
+- Added `RbacCacheService` in `src/rbac/rbac-cache.service.ts` which uses `ioredis` for caching.
+- Integrated a bounded memory cache in `RbacCacheService` and established Redis pub/sub mechanism to instantly propagate invalidation to all instances upon role or permission mutations.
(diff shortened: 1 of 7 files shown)


RUKAYAT-CODER:

PR #1214, opened 2026-08-20
audit_logs: entity/migration disagree on http_method type, version, a…

### LINKED ISSUE #1203: audit_logs: entity/migration disagree on http_method type, version, and timestamp
Migration `1762000000000-create-audit-log-table.ts` creates `http_method` as `varchar` and includes a `version` column; the entity `AuditLog` declares `http_method` as an enum (`audit_logs_http_method_enum`) and has no `version` column. The `timestamp` column is also declared as `timestamptz` in the migration but mapped via `@CreateDateColumn({ name: 'timestamp' })` (timestamp without tz).

Result: schema drift wants to change `http_method` type, drop `version`, and change `timestamp` (all part of the #1194 snapshot).

**Fix:** pick canonical definitions — enum vs varchar for `http_method`, keep or drop `version` — and align the migration with the entity (or add a follow-up migration that reconciles them).

MaryammAli:
## Summary

This PR resolves the schema mismatch between the `AuditLog` entity and the `create-audit-log-table` migration.

## Changes Made

* Aligned the `http_method` definition between the entity and database migration.
* Reconciled the `version` column so the entity and migration use the same schema.
* Aligned the `timestamp` column to use the same timezone-aware definition across the entity and migration.
* Ensured the migration and entity represent a consistent canonical schema to prevent unnecessary schema drift.

## Problem

The migration defined `http_method` as `varchar` and included a `version` column, while the `AuditLog` entity defined `http_method` as an enum and did not include `version`.

Additionally, the migration defined `timestamp` as `timestamptz`, while the entity mapping did not explicitly use the timezone-aware type.

These differences caused schema synchronization to attempt changes to the existing audit log table.

## Result

The `AuditLog` entity and database migration are now consistent, preventing schema drift and ensuring the audit log table is created and maintained with the expected column types and definitions.

Closes #1203

### DIFF
--- a/src/audit-log/audit-log.entity.http-method.spec.ts
+++ b/src/audit-log/audit-log.entity.http-method.spec.ts
@@ -1,0 +1,141 @@
+import 'reflect-metadata';
+import { HttpMethod, AuditLog } from './audit-log.entity';
+
+// ── HttpMethod enum contract ──────────────────────────────────────────────────
+
+describe('HttpMethod enum', () => {
+  it('should contain exactly the five standard HTTP verbs', () => {
+    const members = Object.values(HttpMethod);
+    expect(members).toEqual(expect.arrayContaining(['GET', 'POST', 'PUT', 'DELETE', 'PATCH']));
+    expect(members).toHaveLength(5);
+  });
+
+  it('should have uppercase string values', () => {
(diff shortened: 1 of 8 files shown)


RUKAYAT-CODER:

PR #1340, opened 2026-08-26
fix(moderation): Enforce pagination bounds on content-reports list endpoint (#1306)

### LINKED ISSUE #1306: Enforce pagination bounds on content-reports list endpoints
## Overview

The list endpoints in `src/moderation/reports/content-reports.controller.ts` accept pagination parameters (page/limit/offset/cursor). Ensure a sane maximum page size is enforced and invalid values are rejected, so a client cannot request an unbounded number of rows (a memory/DoS and performance risk) or pass negative/non-numeric values.

## Specifications

Features:

- Page size is clamped to a documented maximum.
- Invalid pagination values (negative, zero, non-numeric, over-max) are rejected with 400.
- A consistent default page size is applied when none is provided.

Tasks:

- Add/replace pagination inputs with a validated pagination DTO (class-validator: `@IsInt`, `@Min`, `@Max`) under `src/moderation/reports`.
- Apply the DTO to the list handlers in `src/moderation/reports/content-reports.controller.ts` and clamp/validate the limit.
- Document the default and maximum page size.

Impacted Files:

- src/moderation/reports/content-reports.controller.ts
- src/moderation/reports/ (pagination DTO)

## Acceptance Criteria

- Requesting an oversized limit is either clamped or rejected with 400 — never served unbounded.
- Invalid pagination values return 400.
- A default page size applies when parameters are omitted.

boalambo:
## Summary
Enforces sane pagination bounds on `src/moderation/reports/content-reports.controller.ts`'s list endpoint, so a client can no longer request an unbounded number of rows (memory/DoS and performance risk) or pass negative/non-numeric pagination values.

Closes #1306

## Changes
- **`dto/list-content-reports-query.dto.ts`**
  - Replaced the unbounded `limit` field with validated `page` and `limit` fields using `@IsInt`, `@Min`, `@Max` from `class-validator`.
  - `limit` is capped at `MAX_PAGE_SIZE` (100) and defaults to `DEFAULT_PAGE_SIZE` (10) when omitted.
  - `page` defaults to `1` and must be a positive integer.
  - Both constants are pulled from the shared `APP_CONSTANTS` for consistency with the rest of the codebase.
  - Added JSDoc documenting the default and maximum page size.
- **`content-reporting.service.ts`**
  - Computes `skip` from `page`/`limit` and passes both `take`/`skip` to the repository query, replacing the old `query.limit ?? 50` fallback that had no upper bound.

## Verification
- `npm run typecheck` passes with no errors.
- Confirmed `ListContentReportsQueryDto` is already wired into the controller's `listReports` handler via `@Query()`, so validation runs through the app's global `ValidationPipe`.
- Confirmed `DEFAULT_PAGE_SIZE` (10) and `MAX_PAGE_SIZE` (100) exist in `src/common/constants/app.constants.ts`.

## Acceptance criteria
- ✅ Requesting an oversized `limit` is rejected with 400 (`@Max(MAX_PAGE_SIZE)`).
- ✅ Invalid pagination values (negative, zero, non-numeric) return 400 (`@IsInt`, `@Min(1)`).
- ✅ A default page size (10) applies when parameters are omitted.

### DIFF
--- a/src/moderation/reports/content-reporting.service.ts
+++ b/src/moderation/reports/content-reporting.service.ts
@@ -80,10 +80,13 @@ export class ContentReportingService {
     if (query.reason) where.reason = query.reason;
     if (query.contentType) where.contentType = query.contentType;
 
+    const skip = (query.page - 1) * query.limit;
+
     return this.reportRepo.find({
       where,
       order: { createdAt: 'DESC' },
-      take: query.limit ?? 50,
+      take: query.limit,
+      skip,
     });
   }
 
--- a/src/moderation/reports/dto/list-content-reports-query.dto.ts
+++ b/src/moderation/reports/dto/list-content-reports-query.dto.ts
@@ -1,9 +1,22 @@
 import { ApiPropertyOptional } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
-import { IsEnum, IsOptional, IsString, Max, Min } from 'class-validator';
+import { IsEnum, IsInt, IsOptional, IsString, Max, Min } from 'class-validator';
+import { APP_CONSTANTS } from '../../../common/constants/app.constants';
 import { ContentReportReason } from '../content-report-reason.enum';
 import { ContentReportStatus } from '../content-report-status.enum';
 
+const { DEFAULT_PAGE_SIZE, MAX_PAGE_SIZE } = APP_CONSTANTS;
+
+/**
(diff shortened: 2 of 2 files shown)


RUKAYAT-CODER:

boalambo:
> Thank you for contributing to the project

Thank you for always, so much grateful. God bless you ma

PR #1345, opened 2026-08-26
fix(courses,tenancy): enforce uniqueness and atomic tenant limits (#1343)

### LINKED ISSUE #1343: Enforce uniqueness and resource limits at the database layer — check-then-act races admit duplicates and over-limit writes
## Overview

Several write paths protect an invariant with an application-level read-then-write (check-then-act) instead of a database constraint or an atomic operation. Under concurrent requests these guards race, admitting duplicate rows or letting counters exceed their limit:

- **Duplicate enrollments.** `enrollments.service.ts` enroll() does `findOne({ userId, courseId })` → `create` → `save`, but `enrollment.entity.ts` has **no unique constraint** on `(userId, courseId)`. Two concurrent enroll requests for the same user+course both pass the "already enrolled?" check and both insert, producing duplicate active enrollments.
- **Duplicate course reviews.** `courses.service.ts` `createReview()` builds and saves a `CourseReview`, and `course-review.entity.ts` only carries a non-unique `@Index()`. A user can create multiple reviews for the same course (concurrently, or simply by calling twice), skewing rating aggregates.
- **Tenant limits exceeded (TOCTOU).** Registration enforces the seat cap by reading `currentUserCount` (`isolation.service.ts` `hasReachedUserLimit()` / `tenant-limit.guard.ts`) and then, separately, calling `tenancy.service.ts` `incrementUserCount()` (an atomic `increment`). Because the check and the increment are separate statements, concurrent registrations can all observe `count < userLimit` and all increment, pushing the tenant over `userLimit` (same shape for `storageLimit`).

The correct pattern already exists elsewhere and should be applied consistently: `forum-vote.entity.ts` uses `@Unique(['entityType','entityId','authorId'])` and `user-quota-usage.entity.ts` uses a unique index plus atomic `increment`. The problem is that these invariants are enforced inconsistently.

## Specifications

Features:

- Uniqueness invariants are guaranteed by database constraints, not only by application-level pre-checks.
- Resource-limit enforcement is atomic with the increment that consumes the resource.
- Constraint violations surface as clean domain errors (409 Conflict) rather than 500s.

Tasks:

- Add a unique constraint on enrollment `(userId, courseId)` (partial/conditional if soft-deleted or non-active rows must be excluded) and rework `enrollments.service.ts` to treat a unique-violation as the authoritative "already enrolled" signal instead of relying on the pre-check.
- Add a unique constraint on course-review `(userId, courseId)` and update `createReview` in `courses.service.ts` to upsert or reject duplicates.
- Make tenant limit enforcement atomic — enforce the cap in the same statement as the increment (e.g. `UPDATE ... SET currentUserCount = currentUserCount + 1 WHERE currentUserCount < userLimit` and treat "0 rows affected" as over-limit) in `tenancy.service.ts`, and reconcile `isolation.service.ts`/`tenant-limit.guard.ts` accordingly.
- Add the backing migration(s) creating the new unique constraints/indexes.
- Add concurrency tests that fire N parallel enroll / review / register requests and assert exactly one write succeeds and limits are never exceeded.

Impacted Files:

- src/courses/enrollments.service.ts
- src/courses/entities/enrollment.entity.ts
- src/courses/courses.service.ts
- src/courses/entities/course-review.entity.ts
- src/tenancy/tenancy.service.ts
- src/tenancy/isolation/isolation.service.ts
- src/tenancy/guards/tenant-limit.guard.ts
- src/auth/auth.controller.ts
- src/migrations/<timestamp>-add-enrollment-and-review-unique-constraints.ts (new migration)
- src/migrations/<timestamp>-enforce-tenant-limits-atomically.ts (new migration)
- test/concurrency-invariants.e2e-spec.ts (new integration test)

## Acceptance Criteria

- Concurrent duplicate enroll requests result in exactly one active enrollment; the second returns the existing one or 409.
- A user cannot have more than one review per course.
- Concurrent registrations can never push a tenant's `currentUserCount`/`currentStorageUsage` beyond its configured limit.
- Unique-constraint violations are mapped to 409 responses, not 500s.
- Concurrency tests cover all three paths and pass.

Nimatstar:
## Summary

Enforces the invariants from #1343 at the database layer:

- **Duplicate enrollment protection**: partial unique index on active enrollments (`UQ_enrollments_active_user_course`) so concurrent/duplicate enroll requests cannot create more than one active enrollment per user+course. Duplicate attempts surface as `409 Conflict`.
- **Atomic tenant limits**: seat consumption (`consumeUserSeat`) and storage usage updates are now atomic conditional updates inside a transaction, so concurrent registrations can never push `currentUserCount` past `userLimit` or overshoot the storage cap.
- **Migration + tests**: new migration `1799000000000-add-enrollment-unique-constraint.ts` and concurrency e2e coverage (`test/concurrency-invariants.e2e-spec.ts`).

All repo CI checks pass locally (lint, typecheck, build, migrations check/run/generate-check/revert).

Closes #1343

### DIFF
--- a/src/auth/auth.controller.ts
+++ b/src/auth/auth.controller.ts
@@ -5,6 +5,7 @@ import {
   UnauthorizedException,
   HttpCode,
   HttpStatus,
+  HttpException,
   Req,
   UseGuards,
   ConflictException,
@@ -19,6 +20,7 @@ import { RegisterDto } from './dto/register.dto';
 import { RefreshTokenDto } from './dto/refresh-token.dto';
 import { InjectRepository } from '@nestjs/typeorm';
 import { User } from '../users/entities/user.entity';
+import { Tenant } from '../tenancy/entities/tenant.entity';
 import { Repository } from 'typeorm';
 import * as bcrypt from 'bcrypt';
 import { JwtAuthGuard } from './guards/jwt-auth.guard';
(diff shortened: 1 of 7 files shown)


RUKAYAT-CODER:

PR #1347, opened 2026-08-27
test(user-preferences): guard currency column against entity drift

### LINKED ISSUE #1206: user_preferences.currency column exists in DB but not in the entity
Migration `1710000000000-AddTimezoneLocalePreferences.ts` adds `currency varchar DEFAULT 'USD'` to `user_preferences`, but `UserPreference` (src/user-preferences/entities/user-preference.entity.ts) only declares `locale`, `language`, `timezone`, etc. — no `currency`.

Same consequences as the `users`/`course` cases: the column can't be read through the entity and the drift check wants to drop it.

**Fix:** declare `currency` on the entity or drop the column.

Ajibose:
Closes #1206

## Summary

`UserPreference` already declares the `currency` column that migration `1710000000000-AddTimezoneLocalePreferences.ts` adds to `user_preferences` (`type: 'varchar', default: 'USD', nullable: true`), matching the migration's `currency varchar DEFAULT 'USD'`. That alignment landed in `79145811` ("fix(migrations): reconcile entity/migration drift and enforce strict schema check"), so the column is already readable through the entity and the drift check no longer wants to drop it.

What was missing was regression coverage: nothing in the test suite would catch `currency` being silently dropped from the entity again in a future change. This PR closes that gap.

### New files
- `src/user-preferences/entities/user-preference.entity.spec.ts` — verifies, via TypeORM's `getMetadataArgsStorage()`, that `UserPreference` declares a `currency` column with `type: 'varchar'`, `default: 'USD'`, `nullable: true` (matching the migration), plus sanity checks for the sibling `locale`/`timezone` columns and that the property itself reads/writes correctly on an instance.
- `src/user-preferences/user-preferences.service.spec.ts` — unit tests (mocked repository, following the existing `preferences.service.spec.ts` pattern) proving `currency` round-trips through `UserPreferencesService#getPreferences`, `#updatePreferences`, and `#resetPreferences`.

### Modified files
None — the entity already declares `currency`; no production code changed.

## Implementation details

- The entity metadata assertion reads TypeORM's internal `MetadataArgsStorage` directly rather than spinning up a real database connection, so the test is fast and has no infra dependency, while still failing loudly if the `@Column` decorator for `currency` is ever removed or its options (`type`/`default`/`nullable`) drift from what the migration produces.
- The service tests mirror the mocking style already used in `src/notifications/preferences/preferences.service.spec.ts` (a plain jest-mocked `Repository`, injected via `getRepositoryToken`) to stay consistent with existing conventions in the codebase.

## Tests added

- `UserPreference entity - schema drift (#1206)`
  - declares a currency column matching the `AddTimezoneLocalePreferences` migration
  - still declares the locale and timezone columns added by the same migration
  - reads a currency value assigned on the entity instance
- `UserPreferencesService - currency (#1206)`
  - returns the currency column when reading an existing preference row
  - persists an updated currency value
  - recreates default preferences with currency undefined so the entity default of USD applies

## How to test

```bash
pnpm install
npx jest src/user-preferences
```

All 6 new tests pass; no other test files were touched.

### DIFF
--- a/src/user-preferences/entities/user-preference.entity.spec.ts
+++ b/src/user-preferences/entities/user-preference.entity.spec.ts
@@ -1,0 +1,29 @@
+import { getMetadataArgsStorage } from 'typeorm';
+import { UserPreference } from './user-preference.entity';
+
+describe('UserPreference entity - schema drift (#1206)', () => {
+  const columns = getMetadataArgsStorage().filterColumns(UserPreference);
+  const columnByProperty = (property: string) =>
+    columns.find((column) => column.propertyName === property);
+
+  it('declares a currency column matching the AddTimezoneLocalePreferences migration', () => {
+    const currencyColumn = columnByProperty('currency');
+
+    expect(currencyColumn).toBeDefined();
+    expect(currencyColumn?.options.type).toBe('varchar');
+    expect(currencyColumn?.options.default).toBe('USD');
+    expect(currencyColumn?.options.nullable).toBe(true);
+  });
+
+  it('still declares the locale and timezone columns added by the same migration', () => {
+    expect(columnByProperty('locale')).toBeDefined();
+    expect(columnByProperty('timezone')).toBeDefined();
+  });
+
+  it('reads a currency value assigned on the entity instance', () => {
+    const preference = new UserPreference();
+    preference.currency = 'NGN';
+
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:
thank you for contributing to the project.

PR #1355, opened 2026-08-27
feat migration

### LINKED ISSUE #1197: course.search_vector: entity column name doesn't match migration or app SQL
Three places disagree on the full-text search column on `course`:

- Migration `1783000000003-add-course-full-text-search.ts` creates `search_vector` (snake_case)
- Entity `Course.searchVector` declares the column as `searchVector` (camelCase, no explicit `name`) with `generatedType: 'STORED'`
- App raw SQL (`src/search/search.service.ts`) queries `course.search_vector` (snake_case)

The entity column maps to a DB column named `searchVector`, which does not exist — the entity can never read the generated column, and the schema drift check wants to drop `search_vector` and create `searchVector`.

**Fix:** declare the entity column with `name: 'search_vector'` (keeping the property name `searchVector`), and align the `asExpression` with the migration's expression. This also removes a chunk of the known drift snapshot in #1194.

Cedarich:
closes #1197


RUKAYAT-CODER:

PR #1370, opened 2026-08-29
fix: Tenant data isolation is not reliably enforced — reads, aggregates, and 

### LINKED ISSUE #1219: Tenant data isolation is not reliably enforced — reads, aggregates, and background jobs bypass RLS
## Overview

`TenantRlsSubscriber` (`src/middleware/tenant/tenant-rls.subscriber.ts`) is the only mechanism enforcing multi-tenant data isolation, and it has several gaps that allow cross-tenant data exposure and integrity violations:

1. **Aggregate/raw queries bypass filtering entirely.** Isolation on reads is done in `afterLoad`, which TypeORM only fires for ORM-hydrated entities. Any `getRawMany()`, `getRawOne()`, `.count()`, or `SUM`/`GROUP BY` query returns rows for **all** tenants unfiltered — e.g. the raw aggregate queries in `audit-log/services/audit-reporting.service.ts` run across every tenant.
2. **Foreign rows are corrupted in place, not removed.** When `afterLoad` detects a row belonging to another tenant it does `Object.keys(entity).forEach(k => k !== 'tenantId' && delete entity[k])`. The row is **still returned** inside the result array (now a partial object with only `tenantId`), so `find()` returns the wrong count and downstream code reads `undefined` fields and crashes instead of the row being excluded.
3. **Background execution has no tenant context, so enforcement is skipped.** `IsolationService` is `@Injectable({ scope: Scope.REQUEST })` and stores the tenant in instance fields set by `tenant.middleware.ts`. Workers, cron jobs, event subscribers, and queue processors run outside a request, so `hasTenantContext()` is `false` and **both** the `afterLoad` filter and the `beforeInsert/Update/Remove` guards no-op — cross-tenant writes and reads proceed unchecked. Injecting a request-scoped service into the globally-registered subscriber is itself unreliable.
4. **Bulk writes bypass the write guard.** `beforeUpdate`/`beforeRemove` read `event.entity`, which is `undefined` for QueryBuilder and `repository.update()/delete()` bulk operations, so those paths skip the cross-tenant check.

Net effect: tenant-scoped tables (`users`, `audit_logs`, CDN content metadata, and the `tenant_config`/`tenant_billing`/`tenant_customization`/`tenant_usage_snapshot` tables) can leak across tenants via raw/aggregate reads, return corrupted partial objects on filtered reads, and accept cross-tenant writes from any non-request execution path.

## Specifications

Features:

- Tenant filtering is enforced at the database layer (Postgres Row-Level Security) so raw, aggregate, and bulk queries cannot cross tenants regardless of the ORM code path.
- Tenant context propagates to non-request execution (workers, cron, subscribers, queue processors) so enforcement is never silently skipped.
- Read isolation removes foreign rows from result sets rather than nullifying their fields.

Tasks:

- Rewrite the `afterLoad` handling so a detected cross-tenant row is treated as an error/excluded rather than mutated in place, and stop deleting entity keys.
- Move tenant context in `isolation.service.ts` from request-scoped instance state to an `AsyncLocalStorage`-backed store so subscribers, workers, and cron see the active tenant; update `tenant.middleware.ts` to populate it.
- Add a migration enabling Postgres RLS and per-tenant policies on the tenant-scoped tables, and set the current tenant on the DB session (GUC, e.g. `SET app.current_tenant`) per connection so filtering happens in the database.
- Apply `applyTenantFilter` (or the DB session tenant) to the raw/aggregate queries in `audit-reporting.service.ts` and to reads/writes in `audit-log.service.ts`.
- Handle bulk `update`/`delete` in the subscriber (or reject untenanted bulk writes on tenant-scoped tables).
- Add integration tests asserting that cross-tenant reads (including `getRawMany` and `.count()`) return nothing, that cross-tenant writes are rejected, and that both hold when executed from a worker with no request scope.

Impacted Files:

- src/middleware/tenant/tenant-rls.subscriber.ts
- src/tenancy/isolation/isolation.service.ts
- src/middleware/tenant/tenant.middleware.ts
- src/middleware/tenant/tenant-access-validation.guard.ts
- src/audit-log/services/audit-reporting.service.ts
- src/audit-log/audit-log.service.ts
- src/cdn/entities/content-metadata.entity.ts
- src/users/entities/user.entity.ts
- src/tenancy/entities/tenant-config.entity.ts
- src/migrations/<timestamp>-enable-tenant-rls.ts (new migration adding RLS policies)
- test/tenant-isolation.e2e-spec.ts (new integration test)

## Acceptance Criteria

- A query for tenant A never returns tenant B's rows, including via `getRawMany`, `getRawOne`, `.count()`, and `SUM`/`GROUP BY` aggregates.
- Filtered reads exclude foreign rows from the result set entirely (correct length, no partial/corrupt objects).
- Cross-tenant inserts, updates, deletes, and bulk writes are rejected.
- Tenant isolation holds for code running outside an HTTP request (workers, cron, subscribers, queue processors).
- Integration tests cover cross-tenant read/write attempts from both request and non-request contexts.

Netty-kun:
## Summary

This pull request implements the changes requested in #1219.

## Changes

- `test/tenant-isolation.e2e-spec.ts`
- `src/middleware/tenant/tenant-rls.subscriber.ts`
- `src/tenancy/isolation/isolation.service.ts`
- `src/middleware/tenant/tenant.middleware.ts`
- `src/audit-log/services/audit-reporting.service.ts`
- `src/audit-log/audit-log.service.ts`
- `src/migrations/1710000000000-enable-tenant-rls.ts`

## Testing

Verified against the issue acceptance criteria.

Closes #1219

### DIFF
--- a/src/audit-log/audit-log.service.ts
+++ b/src/audit-log/audit-log.service.ts
@@ -1,4 +1,5 @@
 import { Injectable } from '@nestjs/common';
+import { IsolationService } from '../tenancy/isolation/isolation.service';
 import { AuditAction, AuditSeverity } from './enums/audit-action.enum';
 import { AuditLog, HttpMethod } from './audit-log.entity';
 import { AuditLoggerService } from './services/audit-logger.service';
@@ -96,12 +97,21 @@ export class AuditLogService {
     private readonly queryService: AuditQueryService,
     private readonly reportingService: AuditReportingService,
     private readonly exportService: AuditExportService,
+    private readonly isolationService: IsolationService,
   ) {}
+
+  private getTenantId(): string {
+    const tenantId = this.isolationService.getTenantId();
+    if (!tenantId) {
+      throw new Error('Tenant context is required for audit operations');
+    }
+    return tenantId;
+  }
 
   // ── Write ──────────────────────────────────────────────────────────────────
 
   log(entry: IAuditLogEntry): Promise<AuditLog> {
-    return this.loggerService.log(entry);
+    return this.loggerService.log({ ...entry, tenantId: this.getTenantId() });
   }
 
   logAuth(options: LogAuthOptions): Promise<AuditLog> {
@@ -113,6 +123,7 @@ export class AuditLogService {
       options.userAgent,
       options.metadata,
       options.severity ?? AuditSeverity.INFO,
+      this.getTenantId(),
     );
   }
 
@@ -127,6 +138,7 @@ export class AuditLogService {
       options.newValues,
       options.ipAddress,
       options.description,
+      this.getTenantId(),
     );
   }
 
@@ -141,6 +153,7 @@ export class AuditLogService {
       options.ipAddress,
       options.userAgent,
       options.requestId,
+      this.getTenantId(),
     );
   }
 
@@ -153,6 +166,7 @@ export class AuditLogService {
       options.userAgent,
       options.description,
       options.metadata,
+      this.getTenantId(),
     );
   }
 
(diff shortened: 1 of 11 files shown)


RUKAYAT-CODER:
Kindly fix workflow to pass.
Run Npm run lint to view and fix errors

RUKAYAT-CODER:
Kindly resolve conflict.
Always pull from the main before pushing your commit

RUKAYAT-CODER:
This PR was closed because of the inability to resolve conflict

PR #1373, opened 2026-08-29
perf(achievements): add lookup indexes

### LINKED ISSUE #1226: Add database indexes to the achievement entity
## Overview

`src/achievements/entities/achievement.entity.ts` declares no `@Index`. Columns that are used for lookups, filtering, sorting, or as foreign keys are therefore unindexed, forcing sequential scans as the table grows. Add appropriate indexes (and a migration that creates them) so the common query paths for this entity stay performant.

## Specifications

Features:

Tasks:

- Review the query paths that hit `achievement` and identify the columns used in WHERE / ORDER BY / JOIN / foreign-key lookups.
- Add `@Index` decorators (single-column or composite as appropriate) to `src/achievements/entities/achievement.entity.ts`.

Impacted Files:

- src/achievements/entities/achievement.entity.ts
- src/migrations/<timestamp>-add-achievement-indexes.ts

## Acceptance Criteria

Bashman10:
## Summary

- Add composite indexes for the active, visible achievement list ordered by difficulty and creation date.
- Add a composite index for active, visible achievements filtered by type and ordered by difficulty.
- Add a reversible TypeORM migration so existing databases receive the indexes.

## Why

These indexes cover the achievement definition lookup paths used by `AchievementsService` without duplicating the primary-key index that already serves ID lookups.

## Verification

- [x] `npm run migrations:check`
- [x] Confirmed the entity decorators and migration create the same named indexes.
- [x] `git diff --check`
- [ ] Full typecheck and Jest suite (blocked locally because the upstream lockfile is out of sync with `package.json`).
- [ ] Live migration run (no PostgreSQL server is available in this environment).

Closes #1226

### DIFF
--- a/src/achievements/entities/achievement.entity.ts
+++ b/src/achievements/entities/achievement.entity.ts
@@ -1,5 +1,6 @@
 import {
   Entity,
+  Index,
   PrimaryGeneratedColumn,
   Column,
   CreateDateColumn,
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:

PR #1377, opened 2026-08-29
fix: remove dead SchemaMigrationService and unused Migration entity

### LINKED ISSUE #1199: SchemaMigrationService and the schema_migrations table are dead code
`src/migrations/schema-migration.service.ts` defines `SchemaMigrationService` which queries/inserts/deletes from a `schema_migrations` table:

- The service is never injected or used anywhere in `src/`
- No migration creates a `schema_migrations` table (migrations use TypeORM's `migrations` table)

It's misleading dead code that also sits in the migrations directory (it had to be excluded from the migration glob in #1192).

**Fix:** remove the file (and `src/migrations/entities/` if also unused), or wire it up and add the table if it's actually needed.

dedukpe:
## Summary

Remove `SchemaMigrationService`, the `Migration` entity, and the `entities/` directory — all dead code that was never wired up.

## Closes

Closes #1199

## Changes

| File | Action |
|------|--------|
| `src/migrations/schema-migration.service.ts` | **Deleted** |
| `src/migrations/entities/migration.entity.ts` | **Deleted** |
| `src/migrations/entities/` | **Deleted** (directory) |
| `src/config/datasource.ts` | Updated comments to reflect the removal |

## Why

`SchemaMigrationService` queries and inserts into a `schema_migrations` table, but:

1. **No module registers it** — it is never injected anywhere in the codebase.
2. **No migration creates the table** — the project uses TypeORM's built-in `migrations` table.
3. **No code imports the `Migration` entity** — zero references across the entire `src/` directory.

The service and entity only sat in the `src/migrations/` directory, where the TypeORM migration glob in `datasource.ts` had to be explicitly configured to ignore them (via the `[0-9]*` timestamp prefix pattern). This makes the code misleading and a maintenance burden.

## Verification

- `tsc --noEmit` passes with no errors
- `nest build` succeeds
- No imports or references to the deleted files remain in the codebase


RUKAYAT-CODER:

PR #1385, opened 2026-08-29
fix: resolve FK dynamically in down() for reversible migration rollback

### LINKED ISSUE #1208: fix-forum-anonymous-author down() isn't cleanly reversible
`1791000000001-fix-forum-anonymous-author.ts`:
- `down()` drops `FK_forum_votes_authorId_users` by name — but that name only exists if the FK was created with it; on dev `synchronize`-created or baseline schemas the FK may have a different name, making `DROP CONSTRAINT IF EXISTS` a silent no-op
- `up()`/`down()` purge/alter `forum_votes` and flag-state changes that can't be reconstructed

This matters now that `migration:run`/`migration:revert` run from scratch in CI (#1193): the revert step must not leave the DB in a broken state.

**Fix:** make `down()` resolve the FK by querying `pg_constraint` (not by hard-coded name), and document the irreversible parts.

Horlarmmy:
## Summary

Fix `down()` in `1791000000001-fix-forum-anonymous-author` to be safely reversible in CI (#1208).

Closes #1208

## Problem

`down()` drops `FK_forum_votes_authorId_users` by hard-coded name. That name only exists if this migration's `up()` created it. On other environments:

| Scenario | Actual FK name | `DROP CONSTRAINT IF EXISTS` result |
|---|---|---|
| After `1796000000001-reconcile-schema-drift` | `FK_930875619d15f219f30923b724c` | Silent no-op |
| TypeORM `synchronize: true` on dev | Auto-generated hex hash | Silent no-op |
| Baseline schema (no FK) | N/A | Silent no-op (ok) |

In the first two cases, `down()` leaves a dangling FK on a column it reverts to `varchar`.

## Fix

- **Dynamic FK resolution**: `down()` now queries `pg_constraint` to find any FK on `forum_votes.authorId -> users.id` by inspecting `pg_get_constraintdef()`, instead of relying on a hard-coded name. If no FK exists, the drop is skipped.
- **Documented irreversible parts**: JSDoc now explicitly notes that purged anonymous votes and `flagged` status changes cannot be reconstructed by `down()`.

## Testing

- `tsc --noEmit` passes
- ESLint passes
- `validate-migrations.js` passes

### DIFF
--- a/src/migrations/1791000000001-fix-forum-anonymous-author.ts
+++ b/src/migrations/1791000000001-fix-forum-anonymous-author.ts
@@ -29,6 +29,17 @@ import { MigrationInterface, QueryRunner } from 'typeorm';
  *  - Converts `forum_votes.authorId` from `varchar` to `uuid` (matching
  *    `users.id`) and adds a foreign key so a synthetic value can never be
  *    written again.
+ *
+ * **Irreversible parts** — `down()` cannot reconstruct:
+ *  - Deleted anonymous vote rows (we don't know how many distinct callers
+ *    voted, only that they collapsed into one row per entityType/entityId).
+ *  - The `status = 'flagged'` update on threads/comments (we can't distinguish
+ *    rows flagged by this migration from rows flagged by auto-moderation).
+ *
+ * **FK constraint name** — `down()` resolves the FK dynamically via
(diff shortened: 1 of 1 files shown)


RUKAYAT-CODER:

PR #1392, opened 2026-08-30
feat: add database indexes to email-template entity

### LINKED ISSUE #1245: Add database indexes to the email-template entity
## Overview

`src/modules/email-template/email-template.entity.ts` declares no `@Index`. Columns that are used for lookups, filtering, sorting, or as foreign keys are therefore unindexed, forcing sequential scans as the table grows. Add appropriate indexes (and a migration that creates them) so the common query paths for this entity stay performant.

## Specifications

Features:

Tasks:

- Review the query paths that hit `email-template` and identify the columns used in WHERE / ORDER BY / JOIN / foreign-key lookups.
- Add `@Index` decorators (single-column or composite as appropriate) to `src/modules/email-template/email-template.entity.ts`.

Impacted Files:

- src/modules/email-template/email-template.entity.ts
- src/migrations/<timestamp>-add-email-template-indexes.ts

## Acceptance Criteria

aliyuHabibu:
Add @Index decorators on the name and createdAt columns of the email_templates entity to keep common lookup and sort query paths performant as the table grows.

Add a migration (IDX_email_templates_name, IDX_email_templates_created_at) so existing databases are updated, matching the entity schema.

The id primary key and key unique constraint are intentionally left unindexed here to avoid duplicate/redundant indexes.

Close #1245


RUKAYAT-CODER:

PR #1400, opened 2026-08-30
feat: add indexes for schema_change entity

### LINKED ISSUE #1238: Add database indexes to the schema_change entity
## Overview

`src/entities/schema_change.entity.ts` declares no `@Index`. Columns that are used for lookups, filtering, sorting, or as foreign keys are therefore unindexed, forcing sequential scans as the table grows. Add appropriate indexes (and a migration that creates them) so the common query paths for this entity stay performant.

## Specifications

Features:

Tasks:

- Review the query paths that hit `schema_change` and identify the columns used in WHERE / ORDER BY / JOIN / foreign-key lookups.
- Add `@Index` decorators (single-column or composite as appropriate) to `src/entities/schema_change.entity.ts`.

Impacted Files:

- src/entities/schema_change.entity.ts
- src/migrations/<timestamp>-add-schema_change-indexes.ts

## Acceptance Criteria

xtrial01:
### Summary
Adds database indexes to the `schema_change` entity to improve query performance, along with the corresponding migration.

### Changes
- Added indexes via decorators on `src/entities/schema_change.entity.ts`
- Added migration `1804000000000-add-schema-change-indexes.ts`

### Verification
- ✅ TypeScript compilation passes (`exit code 0`)
- ✅ Migration timestamp `1804000000000` confirmed unique, no collisions with existing migrations
- ⚠️ Migration test suite fails, but due to a **pre-existing issue unrelated to this change**: two other migrations share duplicate timestamps (`1800000000000` and `1802000000000`). This was verified independently of my changes and should be tracked/fixed separately.

### Notes
The migration test failure predates this PR and isn't caused by the new migration. happy to open a follow-up issue for the duplicate-timestamp cleanup if that's useful.

---
closes #1238

### DIFF
--- a/src/entities/schema_change.entity.ts
+++ b/src/entities/schema_change.entity.ts
@@ -1,4 +1,11 @@
-import { Entity, PrimaryGeneratedColumn, Column, ManyToOne, CreateDateColumn } from 'typeorm';
+import {
+  Entity,
+  PrimaryGeneratedColumn,
+  Column,
+  ManyToOne,
+  CreateDateColumn,
+  Index,
+} from 'typeorm';
 import { SchemaVersion } from './schema_version.entity';
 
 export enum ChangeType {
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:

PR #1409, opened 2026-08-31
fix(caching): bound and prioritise cache warming repository reads

### LINKED ISSUE #1053: Bound the cache warming repository reads in CacheWarmingService
## Overview

`src/caching/cache-warming.service.ts` issues five repository reads (lines 55, 85, 92, 132, 139) against the courses and users tables with no `take`. Cache warming runs on a schedule and on startup, so at each invocation the service loads a full table into memory in order to populate the cache with entries that, by definition, have a bounded useful size. On a large dataset a warming pass can consume more memory than the cache it is filling, and because it runs at startup it competes with the application's own initialisation for both memory and database connections.

## Specifications

**Features:**

- Warming reads a bounded, prioritised working set.
- Warming is throttled so it does not compete with live traffic.

**Tasks:**

- Add an explicit `take` to each read, sized from a configurable `CACHE_WARM_MAX_ENTRIES`.
- Prioritise the warmed set by a real signal — enrollment count for courses, recent activity for users — rather than reading everything.
- Process the working set in batches with a small delay between them to bound database and memory pressure.
- Emit metrics for entries warmed and warming duration.
- Defer startup warming until after the server is accepting traffic.

**Impacted Files:**

- `src/caching/cache-warming.service.ts`
- `src/caching/cache-warming.scheduler.ts`

## Acceptance Criteria

- No warming query reads an unbounded result set.
- The warmed set is selected by a documented priority signal.
- Warming duration and entry count are exported as metrics.
- Startup warming does not delay readiness.

chainsmt:
Bounds all cache-warming repository reads and prevents startup warming from delaying readiness.

- Cap every warming query with an explicit `take` sized from a configurable `CACHE_WARM_MAX_ENTRIES` (ConfigService/env, default 100); replaces the unbounded published-courses find that loaded the whole table into memory.
- Select the warmed set by a real priority signal: published courses ranked by enrollment count (`createdAt` tie-breaker), users by most-recent activity (`lastLoginAt DESC`).
- Process the working set in bounded batches (`CACHE_WARM_BATCH_SIZE`, default 25) with a short delay between batches (`CACHE_WARM_BATCH_DELAY_MS`, default 50ms) to bound DB/memory pressure.
- Export warming metrics: `cache_warm_entries_total`, `cache_warm_last_entries`, and `cache_warm_duration_seconds`.
- Defer startup warming from a blocking `OnModuleInit` to `OnApplicationBootstrap` dispatched via `setImmediate` (not awaited), so warming runs after the server is accepting traffic and never delays readiness.

## Testing
- `npx tsc --noEmit` — clean
- `npm run lint` — clean
- `npx jest src/caching/cache-warming` — 2 suites, 10 tests pass (bounding cap, enrollment ordering, batching/delay, metrics, deferred non-blocking bootstrap). Pre-existing unrelated caching spec failures are unchanged.

Closes #1053

### DIFF
--- a/src/caching/cache-warming.scheduler.spec.ts
+++ b/src/caching/cache-warming.scheduler.spec.ts
@@ -38,8 +38,17 @@ describe('CacheWarmingScheduler', () => {
     );
   });
 
-  it('runs full warm-up on module init', async () => {
-    await scheduler.onModuleInit();
+  it('defers full warm-up to after bootstrap without blocking readiness', async () => {
+    scheduler.onApplicationBootstrap();
+
+    // Warming is dispatched via setImmediate, so it must not run synchronously
+    // during bootstrap.
+    expect(warming.warmAll).not.toHaveBeenCalled();
+
+    await new Promise((resolve) => setImmediate(resolve));
+    // Allow the swallowed promise chain inside runWarmUp to settle.
+    await Promise.resolve();
+
     expect(warming.warmAll).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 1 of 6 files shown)


RUKAYAT-CODER:

PR #1410, opened 2026-08-31
fix(achievements): correct batch notification date filter and bound processing

### LINKED ISSUE #1391: Batch achievement notifications query never matches (uses unlockedAt = now, ignores today boundary)
## Overview

`AchievementsNotificationsService.sendBatchNotifications()` in `src/achievements/achievements-notifications.service.ts` is meant to send notifications for achievements unlocked today. It computes a midnight boundary:

```
const today = new Date();
today.setHours(0, 0, 0, 0);

const achievements = await this.userAchievementRepository.find({
  where: {
    unlockedAt: new Date(),        // <-- exact current timestamp
    notificationSent: false,
  },
  relations: ['user', 'achievement'],
});
```

There are two problems:

1. **The filter never matches.** `unlockedAt: new Date()` compares `unlockedAt` for exact equality with the current instant (millisecond precision), so it will essentially never match any stored row. The computed `today` boundary is never used — the intent was clearly a range filter such as `MoreThanOrEqual(today)`.
2. **The query is unbounded.** Unlike `resendFailedNotifications()` (which uses `take: 100` to batch), this query has no limit, so once the filter is fixed it could load an unbounded number of rows and send them all in a single tight loop.

The net effect is that scheduled batch achievement notifications are silently never sent.

## Specifications

Features:

- The batch job selects achievements unlocked within the intended window (e.g. since midnight today) that have not yet been notified.
- The batch is processed in bounded chunks.

Tasks:

- Fix the `where` clause to use a range predicate against the computed boundary (e.g. `unlockedAt: MoreThanOrEqual(today)`) instead of `unlockedAt: new Date()`.
- Add a `take` limit (and loop/paginate) so the batch is processed in bounded chunks like `resendFailedNotifications()`.
- Add a unit test asserting that an achievement unlocked earlier today with `notificationSent = false` is selected and notified.

Impacted Files:

- src/achievements/achievements-notifications.service.ts
- src/achievements/achievements-notifications.service.spec.ts (new/updated test)

## Acceptance Criteria

- Achievements unlocked during the intended window with `notificationSent = false` are selected and notified.
- The batch query is bounded and processes large backlogs in chunks.
- A regression test covers the corrected date filter.

ZuLu0890:
## Summary

Closes #1391

`AchievementsNotificationsService.sendBatchNotifications()` was silently **never sending** scheduled batch notifications for achievements unlocked today. The query filtered on `unlockedAt: new Date()`, which compares against the exact current instant with millisecond precision and essentially never matches any stored row — the computed midnight boundary (`today.setHours(0, 0, 0, 0)`) was calculated but never used. The query was also unbounded: unlike `resendFailedNotifications()`, it had no `take` limit, so a fixed query would load an unbounded number of rows and send them all in a single tight loop.

### `src/achievements/achievements-notifications.service.ts`
- **Fixed the date filter:** replaced `unlockedAt: new Date()` with `unlockedAt: MoreThanOrEqual(today)`, selecting achievements unlocked since midnight today that have not yet been notified (`notificationSent: false`). The intended "unlocked today" window is now actually applied.
- **Bounded the batch:** added `BATCH_SIZE = 100` and process the backlog in bounded chunks using `take`/`skip` pagination, so large backlogs are handled chunk-by-chunk instead of in one unbounded query — consistent with the existing `resendFailedNotifications()` behavior.
- **Loop hygiene:** the pagination loop terminates when a chunk comes back empty or smaller than the batch size, and is lint-compliant (`while (keepProcessing)` rather than a literal constant condition).

### `src/achievements/achievements-notifications.service.spec.ts`
- **Regression test for the corrected date filter:** asserts the query is issued with `unlockedAt: MoreThanOrEqual(today)` and `notificationSent: false`, and that an achievement unlocked earlier today is selected and notified.
- **Bounded batching test:** verifies a 103-record backlog is processed in chunks (100 + 3) and that all 103 records are marked as sent.

## Verification

All relevant checks run locally:

| Check | Result |
| --- | --- |
| `pnpm lint:ci` | ✅ 0 errors (only pre-existing warnings in unrelated files) |
| `pnpm format:check` | ✅ pass |
| `pnpm typecheck` | ✅ pass |
| `pnpm test src/achievements/achievements-notifications.service.spec.ts` | ✅ 10/10 pass |

Note: `src/achievements/achievements.service.spec.ts` has failing tests, but they fail identically on `main` (unmodified baseline) and are unrelated to this change.

### DIFF
--- a/src/achievements/achievements-notifications.service.spec.ts
+++ b/src/achievements/achievements-notifications.service.spec.ts
@@ -1,5 +1,6 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { getRepositoryToken } from '@nestjs/typeorm';
+import { MoreThanOrEqual } from 'typeorm';
 import { AchievementsNotificationsService } from './achievements-notifications.service';
 import { UserAchievement } from './entities/user-achievement.entity';
 import { Achievement } from './entities/achievement.entity';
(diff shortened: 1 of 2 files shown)


RUKAYAT-CODER:

